PR #23, opened 2014-06-11
Rename endpoints to resources

pedro:
The first is a bit confusing because each file is actually a
collection of endpoints. I like "resources" because it tells devs
what each file should expose, not to mention it sounds more
familiar to API developers and all.

Thoughts @geemus @mfine @brandur ?

### DIFF
--- a/lib/pliny/commands/generator.rb
+++ b/lib/pliny/commands/generator.rb
@@ -27,10 +27,10 @@ module Pliny::Commands
       end
 
       case type
-      when "endpoint"
-        create_endpoint(scaffold: false)
-        create_endpoint_test
-        create_endpoint_acceptance_test(scaffold: false)
+      when "resource"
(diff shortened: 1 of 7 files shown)


gudmundur:
@pedro Isn't this a little too late by now? :)

dmathieu:
I agree it's too late to make such a change.
Also, it would need a major release, probably with deprecation warnings first.

geemus:
Could be ok if it supports both, but might indeed end up in some weird cases of people unexpectedly having both directories and finding that confusing.

pedro:
Right on, I still think it's a good chance but don't care enough to want to drive it :} Closing!

Small note on major release though: being still on major version zero we're still allowed breaking changes, at least in theory :}

gudmundur:
@pedro haha :)

PR #81, opened 2014-10-15
Save the Sequel database instance for seeds

pedro:
This way you can refer to it to insert/modify rows, etc.

(Alternatively you can still require your app initializer and access your models/etc)


gudmundur:
I believe this can be closed out, as https://github.com/interagent/pliny/pull/223 solved this issue.

pedro:
:+1:!

PR #83, opened 2014-10-17
Split README and template README.

jroes:
Here's my attempt at a fix for interagent/pliny-template#140


gudmundur:
@jroes Does this still apply after the template was moved inside of this repository? If it still does, are we in a better position now to resolve this?

PR #95, opened 2014-11-06
Use minitest assertions instead of rspec expectations

pedro:
I prefer the assertion style because:
- It's often shorter
- It usually reads better (eg: `assert_equal 1, foo` vs `expect(foo).to eq(1)`
- I always have to refer to the rspec documentation for the different things you can do with expect, whereas assertions are a standard shared across different frameworks and in some cases languages too

On the other hand I do appreciate that `expect` makes it obvious what is the actual value and expected one, whereas with `assert` you have to know the convention. I eventually got over that though, because it's a really simple standard to learn and it applies everywhere. We should also consider Pliny gives you sample tests with assertions so this doesn't seem relevant here.

Thoughts?

Corresponding pull in the template:
https://github.com/interagent/pliny-template/pull/149

### DIFF
--- a/lib/template/spec/spec_helper.rb
+++ b/lib/template/spec/spec_helper.rb
@@ -36,6 +36,7 @@ RSpec.configure do |config|
     DatabaseCleaner.clean
   end
 
+  config.expect_with :minitest
   config.run_all_when_everything_filtered = true
   config.filter_run :focus
 


pedro:
Should be captured in the discussion above: it's often shorter, reads better, and you don't have to learn 20+ ways to assert stuff (cover, be_within, contain_exactly, meeh)

PR #110, opened 2014-12-20
Fix exception handling with respect to Rollbar

brandur:
Makes some minor fixes to get this sytem working:

* Blank out `sinatra.error` when catching a Pliny exception so that
 Rollbar doesn't see it.
* Add Rollbar Sinatra middleware.
* Remove base endpoint error handling block for Pliny errors (everything
 gets sent up to the middleware).
* Use Rollbar's new `disable_monkey_patch` config as written by Pedro.
* Bump Rollbar version to get new config.

### DIFF
--- a/lib/pliny/middleware/rescue_errors.rb
+++ b/lib/pliny/middleware/rescue_errors.rb
@@ -8,6 +8,9 @@ module Pliny::Middleware
     def call(env)
       @app.call(env)
     rescue Pliny::Errors::Error => e
+      # blank out this field so that the error is not picked up by Rollbar
+      env["sinatra.error"] = nil
+
       Pliny::Errors::Error.render(e)
     rescue Exception => e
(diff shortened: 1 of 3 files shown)


pedro:
@brandur hey! damn this is getting super close, but still seeing `Sinatra::NotFound` errors going to Rollbar!

This is a bit annoying... I wouldn't expect the `not_found` block in Sinatra to leave an exception in `sinatra.error`, but that being the case should we just patch the Rollbar middleware to ignore it?

brandur:
@pedro Doh! Good point.

Ugh, yeah, I'd probably scrub those in the same way that we scrub existing Pliny errors. It's a little janky whitelisting Sinatra errors that are non-exceptional compared to those that actually are, but checking the source it seems like that might be the only special exception that the framework throws.

PR #113, opened 2015-01-05
re-add require of pliny/version

nzoschke:
I accidentally removed it in b3d41186


brandur:
@nzoschke Thanks! Isn't the `VERSION` reference only needed in the `Rakefile`? (and I think it's already included there)

nzoschke:
Without this, `pliny-new` doesn't work.

brandur:
Ah, I see. Thanks Noah! Will release this as 0.6.1.

PR #116, opened 2015-01-06
Ensure we seed when running db:setup

neilmiddleton:
We're not running `db:seed` after setup at the moment.  We can add this to `bin/setup`, or make this change here.


tmaier:
Makes sense. Thank you.

:+1: from me

brandur:
+1! Pulling this in.

PR #122, opened 2015-01-27
Check media_type instead of content_type.

dpiddy:
This will allow Content-Types header values such as "application/json;charset=utf-8".

See [the rack docs](http://www.rubydoc.info/github/rack/rack/Rack/Request#media_type-instance_method).


brandur:
One million times better. Thanks!

PR #123, opened 2015-01-28
Default timeout

dmathieu:
The current default timeout in a generated app is of 45 seconds.
This can generate H12s very easily, and a request should be refactored way before it uses that long.

This PR changes the default value of timeout to 10 seconds.


brandur:
@dmathieu Interesting! This is aggressive, but probably a better default for most use cases. Thanks!

dmathieu:
FTR, 10 seconds is the default in rack-timeout, and what support recommends to all customers.

PR #124, opened 2015-01-29
Alternatively support the `X-` prefix on the `Request-Id` field

brandur:
So that Pliny apps can play nicely with Hermes out of the box [1].

[1] https://devcenter.heroku.com/articles/http-request-id


pedro:
+1!

PR #125, opened 2015-01-29
Remove pry-doc gem because it is 15MB

will:
15 megabytes is a lot of megabytes


brandur:
Haha, fair enough. I must admit that I haven't ended up using this as much as I thought I would.

PR #128, opened 2015-02-12
Bump prmd.

dpiddy:
Mainly to get the fix in https://github.com/interagent/prmd/pull/143.

Also has the generator specs simulate `pliny-new` before running as schema rebuilding has a dependency on meta.json existing. Maybe a bit heavy-handed?


brandur:
Thanks!

> Also has the generator specs simulate pliny-new before running as schema rebuilding has a dependency on meta.json existing. Maybe a bit heavy-handed?

Let's try it :) It seems like we've needed a more reliable way to accurately test the generator anyway, and this might be it.

dpiddy:
A release including this would be much appreciated.

brandur:
@dpiddy Sorry for the delay! Released as 0.6.3.

PR #131, opened 2015-02-18
Just run the base 404 responder on Sinatra::NotFound.

dpiddy:
`not_found` will run the handler on both `Sinatra::NotFound` and any response with status 404. This means if a route purposely 404s (probably with a helpful error message) the block passed to `not_found` will overwrite it.


brandur:
Excellent point. Sometimes the behavior of these Sinatra special helpers can be a little opaque in terms of how they behave anyway. Thanks!

PR #134, opened 2015-03-09
Schema bits live under ./schema

dpiddy:
`bundle exec rake schema` currently drops `schema.json` under `docs`, the [layout suggested by prmd](https://github.com/interagent/prmd#file-layout) has it under `docs/schema`.

This could be a breaking change, thoughts on handling that? Maybe if `docs/schema.json` exists still rebuild to there but print a warning?


brandur:
> bundle exec rake schema currently drops schema.json under docs, the layout suggested by prmd has it under docs/schema.

I'm actually kind of a fan of just having it under `./schema/schema.json` too because it's used for purposes that are _not_ documentation too.

This could be a breaking change, thoughts on handling that? Maybe if docs/schema.json exists still rebuild to there but print a warning?

I like this. Some kind of basic detection mechanism to help users who are on the previous structure.

dpiddy:
Would you put everything under `docs/schema` under `schema` too? So something like:

```
schema/
  meta.json
  schema.json
  schema.md
  schemata/
    foo.yaml
    bar.yaml
```

brandur:
@dpiddy Yep :) Check API's schema directory for example.

dpiddy:
Ok, got this changed up to create new apps with the `schema/` layout above but continue using the legacy layout with a warning if `docs/schema.json` or `docs/schema/schemata` exist.

brandur:
So awesome :) Thanks @dpiddy!

PR #135, opened 2015-03-17
Add 429 Too Many Requests response code

gudmundur:
Noticed that `429 Too Many Requests` was missing from the error codes supported by Pliny.


brandur:
Hah! I think we were missing this because Excon didn't have it for some reason at the time (and IIRC, it was recently added there as well).

Thank-you as usual @gudmundur!

PR #137, opened 2015-03-19
Add default log context

gudmundur:
Ability to set default log contexts that are not tied to requests. This allows an initializer to set `{ app: 'my-app' }` which will be picked up by `Pliny::Extensions::Instruments`. The `Instruments` are currently invoked before the `RequestStore` is seeded.

### DIFF
--- a/lib/pliny/log.rb
+++ b/lib/pliny/log.rb
@@ -1,7 +1,7 @@
 module Pliny
   module Log
     def log(data, &block)
-      data = log_context.merge(local_context.merge(data))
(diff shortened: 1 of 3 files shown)


brandur:
@gudmundur Nice! I love it. Do you think we should put a default initializer for this in the template?

gudmundur:
Yes, I'll add one up in the morning.

gudmundur:
@brandur there you go!

brandur:
@gudmundur Sweet! Thanks again man :)

PR #138, opened 2015-03-19
Use PLINY_ENV over RACK_ENV

neilmiddleton:
`RACK_ENV` shouldn't be used to determine the environment past `development` or `deployment`.

See http://www.hezmatt.org/~mpalmer/blog/2013/10/13/rack_env-its-not-for-you.html

In fact, it can even cause some issues if your app is using unicorn, as they're including internal rack middlewares relying on RACK_ENV to have the deployment value.
See https://github.com/defunkt/unicorn/blob/master/lib/unicorn.rb#L56-L79

Rack itself does that too.
https://github.com/rack/rack/blob/4e4ab39b0508aa3e59f5d7e53696ef6ae7c220ed/lib/rack/server.rb#L228

This PR adds `PLINY_ENV` for dictating environment, and leaves `RACK_ENV` as it should be.

(note there are some failing tests, but I have no idea why these are happening.  Could do with some help)

### DIFF
--- a/lib/template/config/config.rb
+++ b/lib/template/config/config.rb
@@ -11,7 +11,6 @@ module Config
 
   # Mandatory -- exception is raised for these variables when missing.
   mandatory :database_url, string
-  mandatory :pliny_env, string
 
   # Optional -- value is returned or `nil` if it wasn't present.
   optional :placeholder,         string
--- a/lib/template/config/config.rb
+++ b/lib/template/config/config.rb
@@ -11,6 +11,7 @@ module Config
 
   # Mandatory -- exception is raised for these variables when missing.
   mandatory :database_url, string
+  mandatory :pliny_env, string
 
   # Optional -- value is returned or `nil` if it wasn't present.
   optional :placeholder,         string
@@ -25,6 +26,7 @@ module Config
   override :puma_min_threads, 1,    int
   override :puma_workers,     3,    int
   override :rack_env,         'development', string
+  override :pliny_env,        'development', string
   override :raise_errors,     false,         bool
   override :root,             File.expand_path("../../", __FILE__), string
   override :timeout,          10,    int
--- a/lib/template/config/puma.rb
+++ b/lib/template/config/puma.rb
@@ -1,6 +1,6 @@
 require "./config/config"
 
-environment Config.rack_env
+environment Config.pliny_env
 port Config.port
 quiet
 threads Config.puma_min_threads, Config.puma_max_threads
--- a/lib/template/spec/spec_helper.rb
+++ b/lib/template/spec/spec_helper.rb
@@ -5,7 +5,7 @@
 #
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 #
-ENV["PLINY_ENV"] = "test"
+ENV["RACK_ENV"] = "test"
 
 require "bundler"
 Bundler.require(:default, :test)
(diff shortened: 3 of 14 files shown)


dmathieu:
:blue_heart:

lib/template/spec/spec_helper.rb
```diff
@@ -5,7 +5,7 @@
 #
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 #
-ENV["RACK_ENV"] = "test"
+ENV["PLINY_ENV"] = "test"
```
brandur: Damn, I think we might need to leave this one as is, or at least set both. The reason this is here is that Sinatra actually checks it as it's being required and tweaks its error behavior based off of it. I can't recall the exact semantics, but I believe setting test allows stack traces to be properly dumped to console while running specs.

brandur: This does go on to reinforce your point though :) Sinatra implicitly modifying its behavior based on `RACK_ENV` is incredibly confusing and very bad.

dmathieu: Let's fix it in sinatra!

lib/template/config/config.rb
```diff
@@ -11,6 +11,7 @@ module Config
 
   # Mandatory -- exception is raised for these variables when missing.
   mandatory :database_url, string
+  mandatory :pliny_env, string
```
brandur: Hm, we should probably decide on `mandatory` _or_ `override` I think.

brandur:
@neilmiddleton Nice! I wondered what best practice was supposed to be around `RACK_ENV`. This seems reasonable to me.

lib/template/spec/spec_helper.rb
dmathieu: See sinatra/sinatra#984

brandur: @dmathieu Hahaha, nice! I suspect that this might be too much of a breaking change for them at this point though (too many test suites out there relying on it), but I hope you land it.

dmathieu: I discussed quickly with @rkh first, who agreed on the change, but for 2.0 only.

brandur: Ah! That makes more sense. Cool.

lib/template/config/config.rb
neilmiddleton: `override` probably makes more sense.  It then behaves the same as `RAILS_ENV` etc.

neilmiddleton:
I've fixed `spec_helper`, the `mandatory` and rebased from upstream master.

brandur:
@neilmiddleton Awesome! Mind putting the `RACK_ENV` back in the template's `spec_helper` too?

neilmiddleton:
Updated (with a rebase somehow coming through too)

brandur:
Cool, thanks!

lib/template/config/puma.rb
```diff
@@ -1,6 +1,6 @@
 require "./config/config"
 
-environment Config.rack_env
+environment Config.pliny_env
```
gudmundur: @brandur This might not have the intended effect, see: https://github.com/interagent/pliny/issues/237#issuecomment-169757105

brandur: Dang.

Unfortunately the originally linked blog post now seems to be gone. At the time I remember understanding what they were saying in principle, but not ever having personally experienced any actual problems from just using `rack_env`.

One possible easy fix: just change the defaults for `pliny_env` and `rack_env` to `production`. Alternatively, make them mandatory configuration so that users have to put something in the env.

brandur: And also: maybe just remove `rack_env` from the configuration template. It doesn't seem like we actually depend on it anywhere anymore.

PR #140, opened 2015-03-23
Timeout postgres queries after Config.timeout

neilmiddleton:
We're having the [HTTP requests timeout after `Config.timeout` seconds](https://github.com/interagent/pliny/blob/master/lib/template/lib/routes.rb#L7).  We should probably make sure that queries aren't running past this time otherwise it's just a waste.

The only issue I can think of here is around migrations, but those seem to [use a different connection](https://github.com/interagent/pliny/blob/master/lib/pliny/tasks/db.rake#L12)

### DIFF
--- a/lib/template/config/config.rb
+++ b/lib/template/config/config.rb
@@ -28,6 +28,7 @@ module Config
   override :raise_errors,     false,         bool
   override :root,             File.expand_path("../../", __FILE__), string
   override :timeout,          10,    int
+  override :database_timeout, 10,    int
   override :force_ssl,        true,  bool
   override :versioning,       false, bool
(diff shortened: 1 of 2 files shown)


brandur:
@neilmiddleton Good call! What do you think about having a separate config for `statement_timeout` and defaulting it to the same value as `timeout`? I kind of like the idea of the default being good for almost everybody, but that they can also be varied independently as needed.

neilmiddleton:
I've added a separate `database_timeout` defaulted to the same value as `timeout`.  If you do nothing it makes sense, but you can still dick with it as much as you want.

brandur:
@neilmiddleton Awesome. Thank-you!

PR #142, opened 2015-03-31
use Rack::Timeout

### LINKED ISSUE #141: Timeout middleware should not exist or at least not be used by default
Because timing out a thread at any random point is inherently unsafe… e.g. causing state corruption in non-atomic changes.

Cf. [rack-timeout's readme](https://github.com/heroku/rack-timeout) and [this detailed explanation by Charles Nutter](http://headius.blogspot.de/2008/02/rubys-threadraise-threadkill-timeoutrb.html).

So timing out dangerously should be an active choice by the developer, and in that case, using rack-timeout is a better option as at least it provides some useful logging (plus special handling for router wait time and post requests, although it can all be disabled if you want simple.)

pedro:
Fix #141.

Current behavior on timeout is to log the corresponding Rack::Timeout::Error exception to Rollbar and render a 500. I think this is a sensible default, and certainly better than the current state where there's apparently no visibility into timeouts. Thoughts?


lib/template/lib/routes.rb
```diff
@@ -4,7 +4,7 @@
   use Pliny::Middleware::CORS
   use Pliny::Middleware::RequestID
   use Pliny::Middleware::RequestStore, store: Pliny::RequestStore
-  use Pliny::Middleware::Timeout, timeout: Config.timeout if Config.timeout > 0
+  use Rack::Timeout if Config.timeout > 0
```
brandur: I think this might be the wrong `Config`.

brandur: Also, @kch, you should consider taking options into the middleware's constructor! That would allow for varying timeout depending on the middleware stack that you're building (compared to a global option).

lib/template/config/initializers/timeout.rb
```diff
@@ -0,0 +1,3 @@
+if Config.timeout > 0
```
brandur: Should we call this file `rack_timeout.rb` maybe? We recently introduced database timeouts as well, so it might help disambiguate.

brandur:
@pedro Nice dude! Looks good to me.

lib/template/lib/routes.rb
kch: @brandur could do; currently all settings are class-level, so it's not a super straightforward change. opened an issue tho.

pedro: What do you mean, which `Config` should this use?

lib/template/config/initializers/timeout.rb
pedro: Good point, I was debating whether all the timeouts should be in the same file but makes more sense to have one per component/dependency instead.

kch:
Mind it that rack-timeout is still dangerous for the same reasons mentioned in #141. (I don't know how dangerous it's for pliny/sinatra/rack/etc, ie. if they have any global mutable state used past a request's lifetime that could go out of sync)

If you're gonna have rack-timeout in there, you probably want to set database and excon timeouts (to something less than rack's) at least, so the major causes of timeout can be handled more appropriately.

lib/template/lib/routes.rb
brandur: @pedro I don't think that `Routes` is inside the `Pliny` namespace (although it probably was at some point) so I think that this will actually grab the obsolete Ruby `Config` module (now `RbConfig`) instead of `Pliny::Config`, which was probably what was intended. I see now that it was wrong before too though, doh.

pedro: But isn't the app config outside the Pliny namespace anyways? eg: https://github.com/interagent/pliny/blob/master/lib/template/config/config.rb

brandur: Oh, actually I forgot that this was renamed to just `Config`. My bad guys! Please disregard.

brandur:
@kch Yeah, good point. See #140 for added DB timeouts. A default initializer for Excon would probably be a good idea too.

pedro:
Cool, merging them!

And +1 for other timeouts, hope we'll continue to help developers get this right – be it with sensitive defaults, be it with documentation. `Rack::Timeout` is just the last layer of defense.

kch:
Yeah any time I see a rack-timeout error in rollbar, it means more specific timeouts need to be set.

PR #143, opened 2015-04-03
bump ruby-version

geemus:


brandur:
Good call! Thanks Wes.

PR #144, opened 2015-04-08
alpha sort generated migration column names

geemus:


brandur:
Major +1. Thanks Wes!

geemus:
Thanks!

PR #145, opened 2015-04-10
Set application_name

pedro:
To help match a process with the corresponding entry in `pg_stat_activity`.

`DYNO` is set by Heroku – if present the process will be identified like `web.1`.

Otherwise it just uses the current process name, which could be as simple as `console`, or complicated as `puma: cluster worker 0: 78289 [myapp]`. This is probably not ideal for many, but at least it sets a precedence so developers can add whatever else is relevant in their deployment setup (ip, instance id, docker container id, etc).

### DIFF
--- a/lib/template/config/initializers/database.rb
+++ b/lib/template/config/initializers/database.rb
@@ -1,4 +1,5 @@
 database_setup_proc = lambda do |conn|
+  # identify postgres connections coming from this process in pg_stat_activity
   process_identifier = ENV["DYNO"] || File.basename($0).gsub(/\W+/, "_")
   conn.execute "SET statement_timeout = '#{Config.database_timeout}s'"
   conn.execute "SET application_name = #{process_identifier}"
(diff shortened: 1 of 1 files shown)


lib/template/config/initializers/database.rb
```diff
@@ -1,3 +1,9 @@
+database_setup_proc = lambda do |conn|
+  process_identifier = ENV["DYNO"] || File.basename($0)
```
brandur: Mind adding a comment on what `File.basename` does?

brandur:
+1! I guess the one consideration is how much Heroku-specific stuff we want to include (dyno is definitely more of a Herokuism even when compared to things like a `Procfile`), but it's not overly harmful.

pedro:
Right on :{ I was wary about a heroku-specific thing too, but it's one of these areas where there's no single standard for us to use. Added some comments, and stripped non-chars from the identifier to avoid pg errors (didn't notice these first time I tried).

brandur:
Thanks @pedro!

PR #147, opened 2015-04-14
Avoid sql injection in db:nuke task

uhoh-itsmaciek:
Not a big deal since this is only self-injection, but also fixes some cases where table names must be quoted (e.g., a table named `"`--which one can create with the SQL `CREATE TABLE """" (a int)`--causes the current task to fail).

### DIFF
--- a/lib/pliny/tasks/db.rake
+++ b/lib/pliny/tasks/db.rake
@@ -31,8 +31,19 @@ namespace :db do
   task :nuke do
     database_urls.each do |database_url|
       db = Sequel.connect(database_url)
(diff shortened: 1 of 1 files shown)


pedro:
Nice! +1 on fixing the injection.

But what do you think of the move from Sequel's `db.tables` to a query on `pg_catalog`? I kinda like the former not only because it's simpler but also because it seems like a good abstraction (if the schema on pg_catalog changes it will still work).

uhoh-itsmaciek:
Obsolete since `db:nuke` has itself been nuked.

PR #148, opened 2015-04-22
Set Sinatra root

pedro:
This way people can use `public/` and `views/` if they do want to.

(I realize Pliny was never intended to become a web framework, but it's awesome to be able to package a front-end JS framework with it, like Ember or Angular).


neilmiddleton:
:heart:

brandur:
+1!

PR #149, opened 2015-04-22
Rename uuid field to just id

pedro:
Calling the primary row identifier "uuid" is weird - that is definitely the column type, but not a good name/label.

This becomes very clear when doing associations: feels weird to call an association column `author_uuid`, for instance.

### DIFF
--- a/lib/pliny/templates/endpoint_scaffold.erb
+++ b/lib/pliny/templates/endpoint_scaffold.erb
@@ -18,19 +18,19 @@ module Endpoints
       end
 
       get "/:id" do |id|
-        <%= field_name %> = <%= singular_class_name %>.first(uuid: id) || halt(404)
+        <%= field_name %> = <%= singular_class_name %>.first(id: id) || halt(404)
         encode serialize(<%= field_name %>)
       end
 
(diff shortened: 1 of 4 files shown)


brandur:
Yeah, +1! Should we do a major version bump or something for compatibility?

pedro:
Good question, semver is pretty clear, but these things all seem like gray area to me with Pliny... This change is backwards compatible in the sense that it won't affect apps relying on Pliny today, but not in the sense that developers relying on the default values might end up with inconsistencies.

So I think it could go either way, but seeing that Pliny is still going through some pretty big changes I think a minor would be enough. Thoughts?

brandur:
Yeah, good point that it won't actually break anyone, but rather leave them with slightly inconsistent schemas.

> Thoughts?

I'm pretty much neutral on this one. I like `id` better, but am also a bit sensitive to alienating existing users. That said, most users are likely to modify a generated schema before running it, so maybe it's not that big of a problem.

pedro:
Cool, +1 merging then. I think a minor version bump will be enough in this case but let me know if you want to go for a major.

PR #150, opened 2015-04-22
Always include rack test methods

pedro:
Just seems to make things so much simpler; in practice I don't see issues in having these defined for model specs.

Thoughts @brandur ?

### DIFF
--- a/lib/pliny/templates/endpoint_acceptance_test.erb
+++ b/lib/pliny/templates/endpoint_acceptance_test.erb
@@ -2,7 +2,6 @@ require "spec_helper"
 
 describe Endpoints::<%= plural_class_name %> do
   include Committee::Test::Methods
-  include Rack::Test::Methods
 
   def app
     Routes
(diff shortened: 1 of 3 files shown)


bjeanes:
Hmm... This seems like it could present some pretty surprising situations. What are you trying to address with this change?

pedro:
Just hate seeing this include in 20 files when I've always had it once in `spec_helper` without issues.

Have you ran into surprises from this before?

bjeanes:
This exactly, no, but plenty of times with implicit or hidden dependencies. Personally, I prefer the explicitness, even if it is a bit repetitive. I'd argue that it's simpler, even if it's more LOC.

brandur:
@pedro I agree with @bjeanes on this one! In general the repetitiveness may seem a little severe, but in my experience this is just one of those areas that you'll just never have to change. The chances are that those redundant includes will sit there without causing any significant harm until the end of time, and during that span they might help someone who's a beginner to the project glean a tiny bit of understanding into how these tests actually work. If you really want to switch out the web testing framework; I guess that's what `sed` is for ;)

+1 if you feel particularly strongly about it though.

pedro:
Cool! I'm generally in for explicit code, but will allow for some flexibility in tests.

That said I definitely don't feel strongly about this, in particular because the code is generated :} Lets keep it in then!

Thanks folks.

PR #153, opened 2015-04-23
Fix db:rollback

pedro:
Noticed `rake db:rollback` was nuking all my tables... I think this code used to work for sequential migrations (eg: `1_create_table_foo`), but wasn't adapted for timestamp based migrations.

This is quite verbose, but looking in [Sequel's source](https://github.com/jeremyevans/sequel/blob/master/lib/sequel/extensions/migration.rb) I didn't find a way to make this cleaner; thought about instantiating the actual `TimestampMigrator` but that seems like an internal class that we shouldn't rely on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ rvm:
   - 1.9.3
 addons:
   postgresql: '9.3'
+before_script:
+  - createdb pliny-gem-test
 sudo: false
 cache: bundler
 notifications:
--- a/lib/pliny/db_support.rb
+++ b/lib/pliny/db_support.rb
@@ -23,6 +23,16 @@ module Pliny
       Sequel::Migrator.apply(db, "./db/migrate")
     end
 
+    def rollback
+      migrations = Dir["./db/migrate/*.rb"].map { |f| File.basename(f).to_i }.sort
+      current = db[:schema_migrations].order(Sequel.desc(:filename)).first[:filename].to_i
+      target = 0 # by default, rollback everything
+      if i = migrations.index(current)
+        target = migrations[i - 1] || 0
+      end
+      Sequel::Migrator.apply(db, "./db/migrate", target)
+    end
+
     def disconnect
       @db.disconnect
     end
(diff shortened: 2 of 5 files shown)


brandur:
> Noticed rake db:rollback was nuking all my tables... I think this code used to work for sequential migrations (eg: 1_create_table_foo), but wasn't adapted for timestamp based migrations.

Hah, sounds about right.

Looks awesome! A couple notes:

1. We may want to just rollback a single step by default. Rollbacks can be pretty constructive, so some measure of conservatism here might be a good idea.
2. More complicated task bodies like this with no tests are starting to make me a bit nervous ;)

pedro:
> We may want to just rollback a single step by default. Rollbacks can be pretty constructive, so some measure of conservatism here might be a good idea.

Hmm this should be rolling back a single step. Do you mean it should have some checks for extra security?

> More complicated task bodies like this with no tests are starting to make me a bit nervous ;)

Oh man, +1. I put this together to continue moving on with my project but it's about time we wrap all of this migration goodness in a proper class with tests. I'll add these next, lets keep this open.

brandur:
> Hmm this should be rolling back a single step. Do you mean it should have some checks for extra security?

Ah! Sorry, I saw this comment in there:

``` ruby
target = 0 # by default, rollback everything
```

And thought that it meant that everything gets rolled back by default. Upon closer inspection though, I see that it should only go one step.

> Oh man, +1. I put this together to continue moving on with my project but it's about time we wrap all of this migration goodness in a proper class with tests. I'll add these next, lets keep this open.

+1!

pedro:
@brandur done, extracted db:migrate and rollback into another class.

Still not happy with these tests, but it's tricky considering we don't control Sequel (eg: I wish the migrator would take Ruby objects instead of a path to look for migrations in the filesystem). If anything this points to two great opportunities to improve Sequel's migrator.

brandur:
@pedro Dude! You freaking rock!

> Still not happy with these tests, but it's tricky considering we don't control Sequel (eg: I wish the migrator would take Ruby objects instead of a path to look for migrations in the filesystem). If anything this points to two great opportunities to improve Sequel's migrator.

Yeah, +1. It'd be nice to have a little more fine-grain control over this type of stuff. This is still a major step in the right direction though IMO.

pedro:
no way, thanks man!

PR #154, opened 2015-04-28
Add helper to measure metrics in logs

ys:
This will permit to instrument code easily.
Following the l2met convention.

```
Pliny.measure(‘something.important’, source: Config.pliny_env) do
  something_important
end
#=> measure#something.important=123 source=production
```

```
Pliny.count(‘code.lines’, 1000, source: Config.pliny_env)
#=> count#code.lines=1000 source=production
```

### DIFF
--- a/lib/pliny.rb
+++ b/lib/pliny.rb
@@ -8,6 +8,7 @@ require_relative "pliny/extensions/instruments"
 require_relative "pliny/helpers/encode"
 require_relative "pliny/helpers/params"
 require_relative "pliny/log"
+require_relative "pliny/measure"
 require_relative "pliny/request_store"
 require_relative "pliny/router"
 require_relative "pliny/utils"
@@ -19,4 +20,5 @@ require_relative "pliny/middleware/versioning"
 
 module Pliny
   extend Log
+  extend Measure
 end
--- a/lib/pliny/measure.rb
+++ b/lib/pliny/measure.rb
@@ -1,0 +1,15 @@
+module Pliny
+  module Measure
+    def measure(name, options = {})
+      start = Time.now
+      return_value = yield if block_given?
+      duration = Time.now - start
+      log({"measure##{name}": (duration * 1000).round}.merge(options))
+      return_value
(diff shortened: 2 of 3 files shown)


pedro:
Nice, this looks like a good start to me. @brandur ?

Weird CI failure, there must be something else going on :{ we'll need to check it out but +1 to ignore for this pull.

ys:
I like the idea of providing easy helpers like this one.

Maybe as opt-in too.

ys:
The error seems to be unrelated at first, can we rerun that one only?

mattreduce:
@ys @pedro I'm interested in this also, the spec failures look like they could do with a re-run (which I can't trigger). The branch could probably benefit from rebasing master in, as well.

ys:
I will take a closer look on monday:) and will try to bring other stuff
from a project

pedro:
@ys not to be a pain but can you also bring support for samples and multiple metric names? :}

eg:
https://github.com/heroku/telex/blob/master/lib/telex/sample.rb
https://github.com/heroku/api/blob/master/lib/api/sample.rb

Was ready to merge, but realized the interface might change!

Thanks,

ys:
Yeah sure

brandur:
Oof, sorry about the delayed response on this one guys.

My one thought here is that the other common type of metric is a `sample` (e.g. number of jobs in your queue). We may want to include that in the default tool set as well. If you like, [you can reference this implementation that I made for API](https://github.com/heroku/api/blob/e7cda3cc409dd9eae2063ed5fb2fa32bedbbd8bd/lib/api/metrics.rb) which we never ended up merging in.

Also, because a `measure` is a type of metric, it might be worth renaming the module to something different. `Metrics`?

jingweno:
@ys what do you think of making it a separate gem (`pliny-log`?) so that [we could pull in to the API](https://github.com/heroku/api/issues/3769#issuecomment-121778103) as the first component?

ys:
Done by @gudmundur

PR #155, opened 2015-04-29
pliny-update

pedro:
Just a first pass at a very simple update flow: make a patch of the diffs in the template, and apply it to the current folder.

Opening just to collect ideas/hopefully have more people test this. Worked fine for me updating from 0.7.3 to 0.8.0 despite some trouble merging changes to `.env.sample`, `.env.test` and `Gemfile`. Good news is we can probably tweak `patch` or  manually handle a lot of the changes on these.

Definitely want to try some different permutations and add find ways to test before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,13 @@ $ foreman run bin/run 'puts "hello world"'  # Run automated code
 
 (hint: don't forget `foreman run` in development)
 
+### Updating
+
+Use `pliny-update` to update the Pliny app in the current directory.
+
+This not only bumps the version dependency, but also applies any changes that happened in the template app (for instance: new initializer, tweaks in the base endpoint, etc).
+
+
 ## Development
 
 Run tests:
--- a/bin/pliny-update
+++ b/bin/pliny-update
@@ -1,0 +1,12 @@
+#!/usr/bin/env ruby
+
+require "optparse"
+require_relative '../lib/pliny/commands/updater'
+
+ARGV.options do |options|
+  opts = {}
+
+  options.parse!
+
+  Pliny::Commands::Updater.run(ARGV.dup, opts)
+end
--- a/bin/pliny-update
+++ b/bin/pliny-update
@@ -4,9 +4,6 @@ require "optparse"
 require_relative '../lib/pliny/commands/updater'
 
 ARGV.options do |options|
-  opts = {}
-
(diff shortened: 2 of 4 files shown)


pedro:
Got surprisingly good results updating Telex, I think this is a promising foundation (not to mention one of the few approaches that will possibly work). Thoughts @brandur ?

brandur:
@pedro Wow, this is a crazy good idea dude! This will be crazy epic if it becomes part of our standard app development flow.

Just a few notes:

1. Do you think we should put together some documentation and/or high-level design comments somewhere?
2. Should we try to add in some basic testing to try to verify some of this behavior. I can see that it would definitely be a little tricky this particular situation.

pedro:
Hey! Absolutely, +1 for both docs and tests. Just wanted to make sure you're kinda in before going too deep – will take a stab at these and ping you again.

brandur:
@pedro Oh definitely, I love the idea. We have a ton of internal Pliny apps out there, and it kills me how it's going to be incredibly difficult to ship them any modernizations that require changes to the app itself. Lemme know how I can help!

pedro:
@brandur interesting, one of the things I noticed testing is that some releases have `v` prepended and some don't, so the diff is not working for all versions:
https://github.com/interagent/pliny/releases

Do you have a preference for `v123` or just `123` for the tag? I'll normalize either way.
Also, any ideas why we're tagging differently? Are you using the rake script to release or is that not working?

pedro:
Ah it's probably just old stuff right? At first I Thought we were tagging differently but I see you also released with `v`. I'll just add a `v` to the old tags ok?

brandur:
> Ah it's probably just old stuff right? At first I thought we were tagging differently, but now I see you also released with v. I'll just add a v to the old tags ok?

+1! I might have released without the script a couple times. My bad! I've been using it more recently.

pedro:
@brandur ok fixed tags, added a little test, I think this is ready to go (as far as first pass/proofing goes heh)

brandur:
@pedro Dude, looks great. +1 from me! I'm excited to see what people do with this.

pedro:
Thanks!

neilmiddleton:
@dmathieu you've done this by hand recently - how's this look?

PR #156, opened 2015-04-29
remove changelog

pedro:
It got outdated.

I like the idea of a changelog, but we should incorporate it to the release process if we're going to stick with it (eg: `rake release` appends the pull requests/asks for a review?)

Dropping in the meantime.


bjeanes:
The projects that have done this the best in the past is to require relevant commits to be prefixed with one of:

* `new: ...`
* `fixed: ...`
* `optimized: ..`
* etc

Then generate a candidate changelog based on commits with that message prefix. Maybe overkill here, but worth thinking about ways to incrementally generate the changelog instead of having it be a blocker for release; it's really hard to look over all the changes in one go and accurately describe what changed.

It'd be nice if there was some automated way to get PR openers to add a changelog entry, actually...

pedro:
I like it!

Maybe for a light-weight approach we can just compile the pull requests? Looking at [their titles alone](https://github.com/interagent/pliny/pulls?q=is%3Apr+is%3Aclosed) you can already have a ok idea of what's going on.

Merging for now, will continue discussing here!

PR #158, opened 2015-04-30
Make Serializers and Mediators namespaces modules

### LINKED ISSUE #157: Mediators and Serializers are classes
I was assuming `Mediators` and `Serializers` are module namespaces. Turns out they're both classes.

¯\\\_(ツ)_/¯

Is there a reason for this? Is user supposed to be able to instantiate them?

rwz:
Fixes #157


pedro:
Yes! Thanks :}

PR #159, opened 2015-05-15
Add console to template Procfile

rwz:
By default Heroku runs default `irb` when you do `heroku run console` on a Pliny app. This PR makes it actually run Pliny console.


brandur:
+1. Thanks!

dpiddy:
:clap:

PR #160, opened 2015-05-15
Set default Rollbar environment to PLINY_ENV

rwz:


brandur:
@rwz What do you think about decoupling this a bit by adding a `ROLLBAR_ENV` configuration setting? The basic idea is that if you need fine-grain control over your Rollbar environment name, then you have it in a way that can be varied independently from `PLINY_ENV`. Thoughts?

rwz:
Honestly I don't even see the reason to have a separate `PLINY_ENV` to begin with. AFAIK it's only used in one place it completely mimics the way `RACK_ENV` works.

Nonetheless, I think Rollbar environment should be set to something by default, and `PLINY_ENV`/`RACK_ENV` are the beast candidates. The end-user still can introduce a separate variable if inclined to do so.

brandur:
> Nonetheless, I think Rollbar environment should be set to something by default, and PLINY_ENV/RACK_ENV are the best candidates. The end-user still can introduce a separate variable if inclined to do so.

Yeah, I dunno. If you even say are trying to build out a high-fidelity staging environment, you'll probably have `RACK_ENV` set to `production` there (so that you can get authentic behavior from Sinatra and other components that observe it), which would lock its Rollbar and your production Rollbar into the same value.

Seeing as the default with no Rollbar env set is to send all errors into one giant pool, it doesn't seem that bad to just set `pliny_env` to some default value and just make it clear that this can easily be changed by just changing that environmental variable.

mattreduce:
@brandur +1 (reviving an old thread here :) and I hadn't considered that use case for having a `RACK_ENV` and `PLINY_ENV` until now. Allows you to even more closely mimic production in other environments (we can't really change the fact so many gems seem to behave differently in and outside of `RACK_ENV == 'production`) while technically naming it some other environment for log/metric sources, error reporting etc. :boom:

brandur:
@mattreduce Ah, +1, that's very true. I'm still not super convinced that the right thing to do is to set `environment` to `PLINY_ENV` though.

@mattreduce @rwz Let me pitch a real-world example. Say you have Hyperion (or Deployhooks or any other Pliny app) running in four places:

1. Production.
2. Staging.
3. A bunch of dev clouds.
4. Locally in development.

I would think that ideally (and correct me here if you think differently), you would like Rollbar exceptions to be labeled with corresponding environments. Maybe something like this:

1. `production`
2. `staging`
3. `cloud-pavel`, `cloud-matt`, `cloud-brandur`
4. `local-pavel`, `local-matt`, `local-brandur`

This allows you to easily distinguish environments in the Rollbar interface.

`PLINY_ENV` is kind of a special variable that allows us to tweak an environment independently of `RACK_ENV`. Mostly notable, it's used to change the environment of Puma, which would allow you to change various options; e.g. changing error reporting in development vs. production.

Here's where the problem crops up though. If we look at our examples above, the `PLINY_ENV` we want to set looks like this:

1. `production`
2. `production`
3. `production`
4. `development`

This is because 1 through 3 are all high-fidelity production-like environments. We don't want our Pumas running in dev mode there. We want them running like they do in real life. The result is that 1 through 3 all get grouped into the same bucket in Rollbar. You cannot distinguish one from the other.

This is why I pitched the idea of a `ROLLBAR_ENV`. You could tweak this to have the same properties as the proposed format above:

1. `production`
2. `staging`
3. `cloud-pavel`, `cloud-matt`, `cloud-brandur`
4. `local-pavel`, `local-matt`, `local-brandur`

Let me know if I'm misunderstanding how you guys are thinking about the situation though! There may be something here that I haven't thought of.

gudmundur:
The pattern I've been using in my Pliny apps is having this as a separate config var for only Rollbar, i.e. `ROLLBAR_ENVIRONMENT`. I don't think we should be conflating `RUBY_`, `RACK_`, `PLINY_`, `ROLLBAR_` environments, even though potentially duplicating values is not the best.

@rwz What do you think?

rwz:
Ok, I agree that having rollbar env being set to `PLINY_ENV` by default doesn't make much sense. I still think there should be at least some default. Making the generator generate something like `ROLLBAR_ENV` would be nice.

gudmundur:
Closing in favor of #258.

PR #162, opened 2015-05-15
Remove pg and sequel from runtime dependencies

### LINKED ISSUE #161: Remove hard dependencies on pg/sequel
Pliny generates a template with `pg` and `sequel` gems by default and that's totally fine. However you can't completely remove these gems from your dependencies because `pliny` gem itself has them in its runtime dependencies. I don't see a reason for that.

User should be able to remove unused DB code from the app.

rwz:
Fixes #161


brandur:
Makes sense to me. Thanks!

pedro:
awesome :+1:

PR #165, opened 2015-05-22
Make Sinatra::NotFound error a warning

rwz:
Right now Rollbar reports this as a critical error by default.


brandur:
I wouldn't think that this should bubble outside of Sinatra in a production environment right? So I guess it's mostly an issue if Rollbar is enabled and the Rack env is development?

PR #166, opened 2015-05-26
Change mount to run on endpoint generator

duduribeiro:
Is the right command on endpoint generator `run` instead `mount`? 

The default template uses `run`, but the display message ask to add `mount`.


duduribeiro:
Sorry.. the `mount` is correct.. 
I used mount outside the `Pliny::Router` :see_no_evil: 
Now it is ok \o

Thanks

PR #168, opened 2015-05-28
Moar db:rollback fixes

### LINKED ISSUE #167: rake db:rollback not working
Appears to be a regression in the latest version (0.8) when running rake db:rollback. 

The migration doesn't actually run for some reason.

I'm trying to get to the bottom of why.

pedro:
- Fix issue rolling back the first migration
- Do not raise when database is empty
- Do not raise when database doesn't have any migrations applied

fix #167

### DIFF
--- a/lib/pliny/db_support.rb
+++ b/lib/pliny/db_support.rb
@@ -30,6 +30,8 @@ module Pliny
     end
 
     def rollback
+      return unless db.tables.include?(:schema_migrations)
+
       migrations = Dir["./db/migrate/*.rb"].map { |f| File.basename(f).to_i }.sort
       current    = db[:schema_migrations].order(Sequel.desc(:filename)).first[:filename].to_i
       target     = 0 # by default, rollback everything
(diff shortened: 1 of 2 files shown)


lib/pliny/db_support.rb
```diff
@@ -25,16 +25,19 @@ def initialize(url)
...
     def rollback
+      return unless db.tables.include?(:schema_migrations)
+      return unless current = db[:schema_migrations].order(Sequel.desc(:filename)).first
+
       migrations = Dir["./db/migrate/*.rb"].map { |f| File.basename(f).to_i }.sort
-      current = db[:schema_migrations].order(Sequel.desc(:filename)).first[:filename].to_i
-      target = 0 # by default, rollback everything
-      if i = migrations.index(current)
-        target = migrations[i - 1] || 0
+      target     = 0 # by default, rollback everything
```
ys: Isn't it safer to rollback only one? A la Rails?

pedro: Oh it's exactly what it does! it's just that it starts assuming it has to rollback to 0, and adjusts as it finds more recent migrations (just below).

ys: hug sorry:D

pedro: no worries! :}

PR #169, opened 2015-06-02
drop db:nuke and reset tasks

pedro:
Nuke is delicate because dropping tables alone is not guaranteed
to clean it all up (eg: user-defined types). Instead we're just
going to recommend users to drop/recreate the entire database.

With this in mind db:reset becomes just setup.


brandur:
+1. Thanks!

PR #172, opened 2015-06-18
Make migration generator name arg more flexible

DyegoCosta:
Migration names are usually big and this makes it easier to write. This might applies to the other generators, I'm suggesting it at least for the migration due its usual name length

[Review on Reviewable](https://reviewable.io/reviews/interagent/pliny/172)

### DIFF
--- a/spec/commands/generator/migration_spec.rb
+++ b/spec/commands/generator/migration_spec.rb
@@ -2,34 +2,18 @@ require 'pliny/commands/generator'
 require 'pliny/commands/generator/migration'
 require 'spec_helper'
 
-describe Pliny::Commands::Generator::Base do
-  subject { Pliny::Commands::Generator::Migration.new(migration_name, {}, StringIO.new) }
-
+describe Pliny::Commands::Generator::Migration do
   describe '#name' do
-    shared_examples 'a migration name' do
-      it 'generates a snakecase name' do
-        assert_equal 'add_column_foo_to_barz', subject.name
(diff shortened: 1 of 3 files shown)


spec/commands/generator/migration_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'pliny/commands/generator'
+require 'pliny/commands/generator/migration'
+require 'spec_helper'
+
+describe Pliny::Commands::Generator::Base do
+  subject { Pliny::Commands::Generator::Migration.new(migration_name, {}, StringIO.new) }
+
+  describe '#name' do
+    shared_examples 'a migration name' do
```
pedro: We tend to avoid these clever but somewhat magical rspec features, at least when the simpler option seem just as readable/terse. Thoughts/Would you mind tweaking this test?

pedro:
I like it! Would certainly like to see this applied all over too. Thoughts @brandur ?

spec/commands/generator/migration_spec.rb
DyegoCosta: Absolutely, I'll simplify it

brandur:
Makes sense to me! Thanks Dyego!

DyegoCosta:
@pedro @brandur May I go ahead and apply this to the rest of the generators?

pedro:
Absolutely, that would be great! Thanks a lot man.

PR #174, opened 2015-06-26
Set updated_at on model creation

gudmundur:
The default behavior of Sequel is to leave `updated_at` as `null`. This causes some issues when sorting models by `updated_at`. This is also to be consistent with ActiveRecord and simplify JSON schemas.


brandur:
+1! Thanks @gudmundur!

PR #175, opened 2015-06-26
Adds a /schema.json endpoint

gudmundur:
Adds a `GET /schema.json` endpoint.

### DIFF
--- a/lib/template/lib/endpoints/schema.rb
+++ b/lib/template/lib/endpoints/schema.rb
@@ -3,8 +3,8 @@ module Endpoints
     get "/schema.json" do
       content_type("application/schema+json")
       response.headers["Cache-Control"] = "public, max-age=3600"
-        raise Pliny::Errors::NotFound.new("Schema")
       unless File.exists?(schema_filename)
+        raise Pliny::Errors::NotFound.new("This application does not have a schema file.")
       end
       File.read(schema_filename)
     end
--- a/lib/template/lib/endpoints/schema.rb
+++ b/lib/template/lib/endpoints/schema.rb
@@ -1,10 +1,11 @@
 module Endpoints
   class Schema < Base
     get "/schema.json" do
-      content_type("application/schema+json")
-      response.headers["Cache-Control"] = "public, max-age=3600"
+      content_type "application/schema+json"
+      headers["Cache-Control"] = "public, max-age=3600"
       unless File.exists?(schema_filename)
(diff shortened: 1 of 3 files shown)


lib/template/lib/endpoints/schema.rb
```diff
@@ -0,0 +1,19 @@
+module Endpoints
+  class Schema < Base
+    get "/schema.json" do
+      content_type("application/schema+json")
+      response.headers["Cache-Control"] = "public, max-age=3600"
+      filename = schema_filename
+      unless File.exists?(filename)
+        raise Pliny::Errors::NotFound.new("Schema")
```
brandur: What do you think about something slightly more descriptive for this one? Maybe something like "This application does not have a schema file."

brandur:
@gudmundur I like it! Thanks!

I guess we should probably try to get at least a basic test in. What do you think?

gudmundur:
@brandur Finally got around to adding specs for this and doing the change you proposed. Additionally I dropped the `config.expect_with :minitest` from `spec_helper.rb` in the template. I'm not sure why minitest is being used by default since RSpec is a required gem. If there is a proper reason for that I can port the spec over to minitest.

pedro:
@gudmundur ugh haha we should elaborate on this, but the biggest reason we use RSpec is for the binary.

You can see a previous discussion here:
https://github.com/interagent/pliny/pull/95

This discrepancy is pretty bad though, last time thinking about this I was considering adding `bin/test` to Pliny (copying the RSpec format in which you can pick a single test and even a single line to run). It's probably a good time to reevaluate minitest-based runners out there (I remember using `turn` years ago).

Anyways – to keep this consistent with the other tests can you keep using minitest assertions? :D

gudmundur:
@pedro I've finally updated the specs to use minitest assertions. It did leave me in a little bind with mocking `File.exists?` and `File.read`. I added `rr` to the template `Gemfile` to be consistent with pliny itself. I'm not entirely sold on this being the right thing, though. Got any ideas?

pedro:
@gudmundur this looks solid :+1:

We're going to stick with rspec mocks though; rr doesn't provide much over it these days, and recently has changed maintainers so dunno if it's still going to be updated regularly. When you have a sec can you update to use it?

Thanks!

gudmundur:
@pedro Done.

pedro:
wohooo thank you!

PR #176, opened 2015-07-02
Update README blurb for sequel-pg

uhoh-itsmaciek:
The current blurb is a cute joke, but it's not going to sway anyone and the jab is a little immature. The blurb I'm suggesting is one [Postgres has been using for years](http://www.postgresql.org/), but it's a little stodgy, so I'm open to other suggestions. Mostly I'd like to see the current wording go.


mattreduce:
:+1: nice

brandur:
Major +1. Thanks Maciek!

pedro:
+999, that was sloppy to say the least.
Thanks Maciek!

Issue #63, opened 2014-08-14
PG::UndefinedFunction: ERROR:  function uuid_generate_v4() does not exist


schneems:
When running db:migrate I get this error

```
PG::UndefinedFunction: ERROR:  function uuid_generate_v4() does not exist
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

Maybe this should be added to the first migration? 

```ruby
    self.execute <<-EOF
      CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
    EOF
```

dmcinnes:
I believe this should be fixed with the latest release

kennyp:
:+1: I think we're good on this one now.

johnbeynon:
I've just tried this with a new 0.7.3 app and it still doesn't deploy to Heroku cleanly - https://gist.github.com/johnbeynon/b687826ab82752bea863

brandur:
@johnbeynon Ugh, I think we need to clarify the docs, but `bin/setup` is more suitable for local development than for platform deployment. One of the things it does is try to bring up a Postgres database for you by attaching to your locally running Postgres daemon (obviously impossible on a dyno).

Try this instead:

```
bundle exec rake schema:load migrate seed
```

johnbeynon:
Yeah, that definitely works. I made the mistake of migrating first but then the uuid function didn't exist for the primary keys.






On Sat, Mar 21, 2015 at 10:53 AM -0700, "Brandur" <notifications@github.com> wrote:












@johnbeynon Ugh, I think we need to clarify the docs, but bin/setup is more suitable for local development than for platform deployment. One of the things it does is try to bring up a Postgres database for you by attaching to your locally running Postgres daemon (obviously impossible on a dyno).



Try this instead:

bundle exec rake schema:load migrate seed

pedro:
`bin/setup` should now work in prod (see #184)
thanks folks!

PR #181, opened 2015-07-10
Deployment notes about initializing the DB

mars:
Tried running `heroku run bin/setup` but fails with the error:

    Sequel::DatabaseConnectionError: PG::ConnectionBad: FATAL:  permission denied for database "postgres"
    DETAIL:  User does not have CONNECT privilege.
    /app/vendor/bundle/ruby/2.2.0/gems/sequel-4.24.0/lib/sequel/adapters/postgres.rb:237:in `initialize'

Found this solution in [an old issue](https://github.com/interagent/pliny/issues/63#issuecomment-84380506).

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,18 @@ $ bundle install
 $ rake
 ```
 
+## Deployment
+
+Before a Pliny app can use the database, the schema must be loaded & then migrated.
+
+Generally after your initial deployment, execute `rake db:schema:load db:migrate` in the environment's console.
+
+On Heroku specifically:
+```
+$ git push heroku master
+$ heroku run rake db:schema:load db:migrate
+```
+
 ## Meta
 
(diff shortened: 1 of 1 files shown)


dmathieu:
You shouldn't execute `schema:load` on your production app. Only `db:migrate`.

brandur:
Oof, we definitely need to be more clear here on when it's appropriate to call `db/setup` and `schema:load` and when it's not.

mars:
I tried running just `heroku run take db:migrate`, but the default uuid function was undefined.

`heroku run rake db:schema:load` fixed that uuid failure.

pedro:
ok `bin/setup` should work in the next release. Thanks folks!

pedro:
fwiw it will continue to run `db:schema:load` as a shortcut to skip some of the migrations, that should be fine in prod too right? @dmathieu

PR #182, opened 2015-07-11
Add Serialize helpers

ys:
With this serialize helpers, we demodulize the endpoint name to imply the serializer class

```
Endpoints::User #=> Serializers::User
```

But you can also specify it for endpoints with custom names.

```
class Endpoints::User::Actions
  serializer 'User'
end

#=> Serializers::User
```

This will dry a bit more the endpoints.


brandur:
Wow, great stuff as usual Yannick!

Our other option may be to force the user the explicitly spell out the serializer class like:

``` ruby
class Endpoints::User::Actions
 serializer Serializers::User
end
```

... which should work as long as we pull the serializers into the app before the endpoints. Thoughts?

/cc @pedro for thinking on this one as well.

pedro:
Yes, +1! Lets pass the class and avoid the extra dependency on `ActiveSupport` then.

gudmundur:
Nice! I agree with Brandur and Pedro, I think we should be explicit here. @ys Would you mind doing that change so we can get a move on this?

ys:
Yeah, sorry forgot all those stuff:) Will redo and move forward

gudmundur:
Closing this PR as #242 was based on this one, and added a few changes. Thanks @ys.

PR #183, opened 2015-07-11
Return an id and message in body when 404 in json.

ys:
All other errors return something with `id` and `message`.
This is just uniformising the behavior there.

### DIFF
--- a/lib/template/lib/endpoints/base.rb
+++ b/lib/template/lib/endpoints/base.rb
@@ -20,7 +20,7 @@ module Endpoints
     error Sinatra::NotFound do
       content_type :json
       status 404
-      "{}"
+      { id: "not_found", message: "Resource not found" }.to_json
     end
   end
 end


brandur:
Nice @ys. Thanks!

pedro:
+1, awesome! can/should we use `encode` there instead of `to_json` though? jumped the eye as I don't think MultiJson defines `to_json`, hmm.

ys:
@pedro I think that it is because pliny has activeSupport:P

But yeah make sense.

bjeanes:
> pliny has activeSupport

It might be healthy to make it `require: false` in Gemfile and just require in the sub-sections of it that are needed, if possible...

pedro:
UGH, great find, I keep forgetting about it... ok I'm taking that out next heh, we only just use it for pluralization.

mattreduce:
@bjeanes hmm yeah I'd say that's an improvement in itself, until AS can be completely removed. Unfortunately the dependency is specified in the gemspec, so we'd need to move it to the `Gemfile` to avoid requiring all of it

pedro:
Opened #186 proposing an alternative to the 404 handler

PR #184, opened 2015-07-13
Make existing db check in rake db:create explicit

pedro:
This way Pliny will also avoid trying to create a db when it doesn't have permissions.

Fix #181


brandur:
Great idea! And awesome to see this one finally fixed. +1 as far as I'm concerned.

pedro:
Thanks!! Merging then.

PR #185, opened 2015-07-14
Do not render status with the error representation

pedro:
That's already in the http status. rendering again is inconsistent
(there's no `status: 200` for successful responses), and just makes
the line between response data and metadata more blur.

Thoughts?


mattreduce:
Makes sense! :100:

ys:
Agree on this one.
A lot of APIs just publish it anyway... but yeah, headers and status for the metadata

pedro:
Thanks folks! Merging.

PR #186, opened 2015-07-14
bridge Sinatra::NotFound to corresponding Pliny::Error

pedro:
This way we're reusing the error serialization and handling defined elsewhere.


ys:
<3

brandur:
Love it. Thanks!

PR #188, opened 2015-07-15
Add generate scaffold integration specs

DyegoCosta:
Given these new specs would hardly break and almost don't add any overhead to the test suite I thould it would be nice to have them

### DIFF
--- a/spec/integration_spec.rb
+++ b/spec/integration_spec.rb
@@ -16,13 +16,25 @@ describe "Pliny integration test" do
     end
   end
 
-  describe "pliny-generate model" do
+  describe "pliny-generate scaffold" do
     before(:all) do
-      bash_app "pliny-generate model artist"
+      bash_app "pliny-generate scaffold artist"
     end
 
     it "creates the model file" do
       assert File.exists?("./myapp/lib/models/artist.rb")
+    end
+
+    it "creates the endpoint file" do
+      assert File.exists?("./myapp/lib/endpoints/artists.rb")
+    end
+
+    it "creates the serializer file" do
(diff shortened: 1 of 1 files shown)


brandur:
@DyegoCosta +1! Do you think we should leave the model path in there as a separate set of specs though? Might be handy to keep that tested as well.

DyegoCosta:
@brandur actually I just realized this turns out to be duplicating the [generator_spec](https://github.com/interagent/pliny/blob/master/spec/commands/generator_spec.rb) specs.
As sanity tests, as the existing model path apparently intends to be, maybe we should add one simple spec for each Thor task. Is it worth it? cc @pedro

brandur:
> @brandur actually I just realized this turns out to be duplicating the generator_spec specs.
As sanity tests, as the existing model path apparently intends to be, maybe we should add one simple spec for each Thor task. Is it worth it? cc @pedro

Oh, great point. Having top-level tests for every task definitely couldn't hurt.

pedro:
Hey!

Right on, this definitely makes this "integration test" a bit more thorough.

Going forward we obviously need a much better story around this, and +1 for testing all the different binaries.

For this top level integration test I was hoping to go from running `pliny-new` to maybe editing an endpoint and running tests on the generated app, but it's a lot of work to do all of that from within a Pliny test... Bundler gets all confused, the bin paths are all weird, etc.

DyegoCosta:
@brandur Just a quick follow up: I tried adding a top-level test for each task and it turns out to increase a lot the suite run time, almost double it. On #191 I tried to make the tasks tinner which might help us to take another approach for testing it.

PR #177, opened 2015-07-02
Make generators name argument more flexible

DyegoCosta:
Complements #172

PR #190, opened 2015-07-22
More strict args in pliny-generate

### LINKED ISSUE #189: Thor required arguments don't seem to be required.
```
$ pliny-generate serializer
created serializer file ./lib/serializers/.rb
created test ./spec/serializers/_spec.rb
```

These commands are Thor actions, which indicates that the name should be required at the Thor level: https://github.com/erikhuda/thor/wiki/Getting-Started#task-parameters

I'm hesitant to add logic here if Thor is supposed to fail the task for us. Thoughts? Could it be because we're not inheriting from Thor correctly? Looking at the Thor docs, it seems correct, but we aren't taking advantage of `Thor::Actions` in the generators class. Doesn't seem like we need it, though.

pedro:
Looking now I think we should not take something like `pliny-generate endpoint my endpoint`.

Handling spaces is convenient, but users should escape that in their shell or opt parsing can get pretty hectic. So `pliny-generate endpoint 'my endpoint'` instead.

See #177. /cc @DyegoCosta

Fix #189

### DIFF
--- a/lib/pliny/commands/generator.rb
+++ b/lib/pliny/commands/generator.rb
@@ -3,9 +3,9 @@ require 'thor'
 
 module Pliny::Commands
   class Generator < Thor
-    desc 'endpoint [NAME]', 'Generates an endpoint'
+    desc 'endpoint NAME', 'Generates an endpoint'
     method_option :scaffold, type: :boolean, default: false, hide: true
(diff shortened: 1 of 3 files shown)


DyegoCosta:
:thumbsup: for escaping the spaces

PR #191, opened 2015-07-22
Extract generators generation steps

DyegoCosta:
This intends to facilitate adding new particularities to each type of generator.

I'm wondering if the `#scaffold` method should get it's own generator and we test only the calls to the other generators. Would that be a good way to go, assuming this refactoring is valid?


gudmundur:
Nice @DyegoCosta. I know it's been ages since you opened this PR, but do you mind rebasing this?

PR #193, opened 2015-07-22
a start at docker support

jroes:
This seems to be working, tests need some cleanup.

Before I do that, though, I'm concerned that this might be too forceful and we'd all have to sort of want to *really* embrace docker-compose here. When I say forceful, I mean:

* The env config has to use postgres hostnames
* The instructions given to users is a docker-compose command

Is there another way to introduce this as slightly more optional, or is that the path interagent wants to take?

Maybe the best way forward is something like `pliny-new --docker myapp`?

```console
~/s/r/pliny ❯❯❯ cd myapp && docker-compose run web bin/setup
Adding DATABASE_URL to .env
Adding PLINY_ENV to .env
Adding TZ to .env
Adding RAISE_ERRORS to .env
Adding FORCE_SSL to .env
Adding TIMEOUT to .env
Adding PRETTY_JSON to .env
Don't run Bundler as root. Bundler can ask for sudo if it is needed, and installing your bundle as root will break
this application for all non-root users on this machine.
Fetching gem metadata from https://rubygems.org/..........
<snip>
Bundle complete! 24 Gemfile dependencies, 56 gems now installed.
Bundled gems are installed into /usr/local/bundle.
Dropped `myapp-development`
Dropped `myapp-test`
Created `myapp-development`
Created `myapp-test`
Loaded `myapp-development`
Loaded `myapp-test`
~/s/r/p/myapp ❯❯❯ docker-compose up
Recreating myapp_postgres_1...
Recreating myapp_web_1...
Attaching to myapp_postgres_1, myapp_web_1
postgres_1 | LOG:  database system was shut down at 2015-07-22 18:30:28 UTC
postgres_1 | LOG:  MultiXact member wraparound protections are now enabled
postgres_1 | LOG:  database system is ready to accept connections
postgres_1 | LOG:  autovacuum launcher started
web_1      | 18:30:53 web.1  | started with pid 8
web_1      | 18:30:53 web.1  | [8] Puma starting in cluster mode...
web_1      | 18:30:53 web.1  | [8] * Version 2.12.2 (ruby 2.2.2-p95), codename: Plutonian Photo Shoot
<snip>
```

### DIFF
--- a/lib/pliny/commands/creator.rb
+++ b/lib/pliny/commands/creator.rb
@@ -5,6 +5,12 @@ require 'uri'
 
 module Pliny::Commands
   class Creator
+    DEFAULT_DB_URI_PREFIX = 'postgres:///'.freeze
+    DOCKER_DB_URI_PREFIX = 'postgres://postgres@postgres/'.freeze
+    # ruby's URI#to_s renders foo:/bar when there's no host
+    # we want foo:///bar instead when there's a single '/'
+    DB_URI_FIX_REGEX = /:\/(?=[^\/])/
+
     attr_accessor :args, :opts, :stream
 
     def self.run(args, opts = {}, stream = $stdout)
@@ -24,13 +30,21 @@ module Pliny::Commands
       FileUtils.rm_rf("#{app_dir}/.git")
       setup_database_urls
       display 'Pliny app created. To start, run:'
-      display "cd #{app_dir} && bin/setup"
+      if opts[:docker]
+        display "cd #{app_dir} && docker-compose run web bin/setup"
+      else
+        display "cd #{app_dir} && bin/setup"
+      end
     end
 
     protected
 
     def setup_database_urls
-      db = URI.parse("postgres:///#{name}")
+      if opts[:docker]
+        db = URI.parse(DOCKER_DB_URI_PREFIX + name)
+      else
+        db = URI.parse(DEFAULT_DB_URI_PREFIX + name)
+      end
       {
         '.env.sample' => 'development',
         '.env.test'   => 'test'
@@ -39,9 +53,7 @@ module Pliny::Commands
         db.path  = "/#{name}-#{db_env_suffix}"
         env      = File.read(env_path)
         File.open(env_path, 'w') do |f|
-          # ruby's URI#to_s renders foo:/bar when there's no host
-          # we want foo:///bar instead!
-          db_url = db.to_s.sub(':/', ':///')
+          db_url = db.to_s.sub DB_URI_FIX_REGEX, ':///'
           f.puts env.sub(/DATABASE_URL=.*/, "DATABASE_URL=#{db_url}")
         end
       end
--- a/lib/template/Gemfile
+++ b/lib/template/Gemfile
@@ -20,6 +20,7 @@ gem "sucker_punch"
 
 group :development, :test do
   gem "pry-byebug"
+  gem "foreman"
 end
 
 group :test do
--- a/lib/template/docker-compose.yml
+++ b/lib/template/docker-compose.yml
@@ -1,0 +1,11 @@
+postgres:
+  image: postgres
+web:
+  build: .
+  command: bundle exec foreman start
+  volumes:
+    - .:/app
+  ports:
+    - "5000:5000"
+  links:
+    - postgres
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 Gemfile.lock
 .bundle
 .ruby-*
+tmp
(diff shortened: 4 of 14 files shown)


markpundsack:
Awesome!

Really wish we didn't need to use foreman. Dropping the database shouldn't be necessary on setup, should it?

I wonder what this would look like if it was based on `heroku docker:init`. Would there be material differences?

jroes:
Happy to wrap this up, just wanted some feedback before continuing. /cc @pedro @brandur

pedro:
YES! Sorry for the delay, this is absolutely awesome.

I agree it's better to leave this optional with `--docker`. Then we can test/announce internally and decide on whether to make it official or not. Would you mind making these changes?

jroes:
Definitely!

jroes:
I've got the flag working, been working on getting the tests to pass (and adding docker-specific ones).

There's some issue with running the container in /tmp, so I had to create a gitignored tmp directory in the project directory.

The main problem I'm running into is that all of the `chdir`ing we're doing is causing the pwd to get out of whack, and generally things are a bit confusing. I tried to reset to the original pwd from the spec helper for all test suites, but it wasn't quite working, so now I'm sprinkling it into `after` blocks whereever we're setting it. It's a bit hard to follow and I haven't finished it up yet.

jroes:
All tests are now passing, rebased, and squashed. Ready for review @pedro @brandur

pedro:
@jroes This looks great to me :+1::+1::+1::+1:
Do you know more about this Travis flag and how can we get in?

markpundsack:
BTW, we got approval from Security to use Circle CI, and they have Docker support. Should we switch?

jroes:
The Travis feature is public now, and it's the same interface I was using for the beta, already in this PR.

If you look at the failure right now, it's something strange with postgres not being installed. Unless someone here knows off the top of their head what's going on there, I might have to ask Travis support folks.

@markpundsack the Travis feature is public now. I'm not sure what the advantages/disadvantages are for Circle, I'm game for whatever.

jroes:
Can't reproduce the Travis error locally. Can anyone else give this a shot?

jroes:
@markpundsack re your earlier comment about `heroku docker:init`, I'm also interested in knowing the differences and seeing if it's something more standardizable.

friism:
@jroes I don't want to be a bigot, but it'd be neat if this used the `heroku/ruby` docker image

markpundsack:
Having just gotten two pliny apps to run on Docker based on `heroku docker:init`, I'd say we should definitely move that direction. At least assuming Pliny apps are aimed at running on the platform. We'd also need to add an app.json file to the Pliny template, which is probably a good idea too.

However for now, I had to use `hone/ruby` rather than `heroku/ruby` so until that is resolved, we're kinda SOL. Personally I say we ship `hone/ruby` asap and continue the debate about dynamic/static ruby version detection later.

lib/pliny/commands/creator.rb
```diff
@@ -24,13 +30,21 @@ def run!
       FileUtils.rm_rf("#{app_dir}/.git")
       setup_database_urls
       display 'Pliny app created. To start, run:'
-      display "cd #{app_dir} && bin/setup"
+      if opts[:docker]
+        display "cd #{app_dir} && docker-compose run web bin/setup"
```
markpundsack: Does this actually work? When I run bin/setup under docker, it complains about dropping the database when there's a current connection. Also, bin/setup does other things like `bundle install` which shouldn't be necessary.

lib/template/Gemfile
```diff
@@ -20,6 +20,7 @@ gem "sucker_punch"
 
 group :development, :test do
   gem "pry-byebug"
+  gem "foreman"
```
markpundsack: Why is foreman necessary?

lib/template/docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+postgres:
+  image: postgres
+web:
+  build: .
+  command: bundle exec foreman start
```
markpundsack: Ah, I see. Do you really want all of the processes run within the web container as opposed to making separate containers for each entry in Procfile?

jroes:
I've run out of time to see this through.

We may want to defer to [heroku-docker](https://devcenter.heroku.com/articles/docker), and in such a case, it may not be necessary to build this directly into Pliny.

From what I can tell, while being initially built at Heroku, Pliny lives a semi-independent existence from Heroku, so I leave this to the maintainers and community around Pliny to make the decisions here.

gudmundur:
@jroes this branch has been stale for a while. I'd love to see this stuff happen though. Is there remaining work on this? If so, what needs to happen?

jroes:
This is way out of date. It'd be interesting to see what Docker support might look like for Pliny, but I doubt this is a useful starting point any longer.

PR #195, opened 2015-07-30
Improve pliny-new command error handling and usage info

Papipo:
I saw that when you ran the generator without arguments, something bad happened:

```
$ pliny-new
/Users/Papipo/.rvm/gems/ruby-2.2.1/gems/pliny-0.10.0/lib/pliny/commands/creator.rb:63:in `initialize': no implicit conversion of nil into String (TypeError)
```

So I have just fixed that and added handling of unknown flags (anyone at the moment).

### DIFF
--- a/bin/pliny-new
+++ b/bin/pliny-new
@@ -3,10 +3,19 @@
 require "optparse"
 require_relative '../lib/pliny/commands/creator'
 
-ARGV.options do |options|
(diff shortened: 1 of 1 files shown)


brandur:
Nice! What we've had up to this point was definitely worrying fragile. Thanks for the contribution @Papipo!

PR #199, opened 2015-08-06
Run test database migrations if schema is outdated

### LINKED ISSUE #197: Running migrations before running tests
It would be lovely if [before running the tests](https://github.com/interagent/pliny/blob/master/spec/spec_helper.rb#L20), the `spec_helper` would ensure all migrations have been ran against the test database. It's a nice way to ensure tests run against the latest schema.

DyegoCosta:
Check if latest migrations were applied before running the test suite.
Apply migrations if schema is outdated.

Resolves #197

[Review on Reviewable](https://reviewable.io/reviews/interagent/pliny/199)

### DIFF
--- a/lib/template/spec/spec_helper.rb
+++ b/lib/template/spec/spec_helper.rb
@@ -31,6 +31,10 @@ Dotenv.load('.env.test')
 
 require_relative "../lib/initializer"
 
+unless Sequel::Migrator.is_current?(Sequel::Model.db, 'db/migrate')
+  Sequel::Migrator.run(Sequel::Model.db, 'db/migrate')
+end
+
 # pull in test initializers
 Pliny::Utils.require_glob("#{Config.root}/spec/support/**/*.rb")
 


uhoh-itsmaciek:
+1! We moved form a bespoke system that ran migrations automatically to Pliny, and having to rerun everything is a pain. Although I've found that rspec can leave a dirty database schema in place if something fails badly enough, so maybe we should just rerun `db:setup` before specs?

cyberdelia:
This also don't restore merged migrations (the ones in ``schema.sql``).

DyegoCosta:
Excellent points. How do you guys feel about adding a flag here to disable or enable that behaviour? Running a full setup might be something people wouldn't want while TDDing

gudmundur:
Personally, I can't say I'm a big fan of automatically running migrations in this case. I think it's fair that we detect this, abort and notify people that they have to run the migrations. What do you think about that approach?

uhoh-itsmaciek:
@gudmundur why the extra step?

PR #202, opened 2015-08-25
Don't load the db tasks if sequel has been removed

dmathieu:
This fixes loading `pliny/tasks` for apps that aren't using a database.


mattreduce:
Nice :+1: I ended up removing these from my Pliny project that doesn't use a database

brandur:
+1. Thanks @dmathieu!

PR #205, opened 2015-09-02
Order overrided config

pedro:
Total cosmetics, but bothered me working with a Pliny app as we had to add several more to the list.


brandur:
+1!!

tmaier:
+1

PR #206, opened 2015-09-02
spec/spec_support

pedro:
Free the namespace `support` to the app (so you can keep other assorted files under `lib/support`.

Also moving the ci/coverage stuff into a new file to keep `spec_helper.rb` lean.


brandur:
Great call. +1.

Issue #207, opened 2015-09-04
Updating the rack-timeout gem break pliny


DyegoCosta:
Using rack-timeout 0.3.1

```
$ rake
rake aborted!
LoadError: cannot load such file -- rack/timeout
/Users/dyego/.rvm/gems/ruby-2.2.3/gems/pliny-0.11.1/lib/pliny.rb:2:in `require'
/Users/dyego/.rvm/gems/ruby-2.2.3/gems/pliny-0.11.1/lib/pliny.rb:2:in `<top (required)>'
/Users/dyego/.rvm/gems/ruby-2.2.3/gems/bundler-1.10.6/lib/bundler/runtime.rb:76:in `require'
/Users/dyego/.rvm/gems/ruby-2.2.3/gems/bundler-1.10.6/lib/bundler/runtime.rb:76:in `block (2 levels) in require'
/Users/dyego/.rvm/gems/ruby-2.2.3/gems/bundler-1.10.6/lib/bundler/runtime.rb:72:in `each'
/Users/dyego/.rvm/gems/ruby-2.2.3/gems/bundler-1.10.6/lib/bundler/runtime.rb:72:in `block in require'
/Users/dyego/.rvm/gems/ruby-2.2.3/gems/bundler-1.10.6/lib/bundler/runtime.rb:61:in `each'
/Users/dyego/.rvm/gems/ruby-2.2.3/gems/bundler-1.10.6/lib/bundler/runtime.rb:61:in `require'
/Users/dyego/.rvm/gems/ruby-2.2.3/gems/bundler-1.10.6/lib/bundler.rb:134:in `require'
/Users/dyego/code/dyego-app/spec/spec_helper.rb:11:in `<top (required)>'
/Users/dyego/code/dyego-app/spec/acceptance/artists_spec.rb:1:in `<top (required)>'
/Users/dyego/.rvm/gems/ruby-2.2.3/gems/rspec-core-3.3.2/lib/rspec/core/configuration.rb:1327:in `load'
/Users/dyego/.rvm/gems/ruby-2.2.3/gems/rspec-core-3.3.2/lib/rspec/core/configuration.rb:1327:in `block in load_spec_files'
/Users/dyego/.rvm/gems/ruby-2.2.3/gems/rspec-core-3.3.2/lib/rspec/core/configuration.rb:1325:in `each'
[...]

PR #208, opened 2015-09-04
Fix loading of rack-timeout

mattreduce:
As of https://github.com/heroku/rack-timeout/commit/6e82c2760723a76f8cf705de6c8d2f588913d94e you must required 'rack-timeout' rather than 'rack/timeout'

Fixes interagent/pliny#207

### DIFF
--- a/lib/pliny.rb
+++ b/lib/pliny.rb
@@ -1,5 +1,5 @@
 require "multi_json"
-require "rack/timeout"
+require "rack-timeout"
 require "sinatra/base"
 
 require_relative "pliny/version"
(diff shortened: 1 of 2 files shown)


tmaier:
Does this also require an update to the dependency in the gemspec file?

mattreduce:
@tmaier ah good point—this should also require at least the version of `rack-timeout` that uses this new file structure.

tmaier:
Thank you!

kch:
Keep in mind that in 0.2.x requiring `rack/timeout` would be equivalent to in 0.3.x requiring `rack/timeout/core`, which does not setup logging.

Requiring `rack-timeout` means requiring `rack/timeout/base`, which is basically core + init logging.

PR #213, opened 2015-09-16
Refactor errors

pedro:
- Constructors take options hash
- Support for metadata (new error attributes that are serialized)
- Automatically render 404s for `Sequel::NoMatchingRow` errors

### DIFF
--- a/lib/pliny/errors.rb
+++ b/lib/pliny/errors.rb
@@ -9,21 +9,21 @@ module Pliny
         [error.status, headers, [MultiJson.encode(data)]]
       end
 
-      def initialize(message, id)
-        @id = id
-        super(message)
+      def initialize(options={})
+        @id = options[:id]
+        super(options[:message])
       end
     end
 
     class HTTPStatusError < Error
       attr_accessor :status
 
-      def initialize(message = nil, id = nil, status = nil)
-        meta    = Pliny::Errors::META[self.class]
-        message = message || meta[1] + "."
-        id      = id || meta[1].downcase.tr(' ', '_').to_sym
-        @status = status || meta[0]
-        super(message, id)
+      def initialize(options={})
+        meta = Pliny::Errors::META[self.class]
+        options[:message] ||= "#{meta[1]}."
+        options[:id] ||= meta[1].downcase.tr(' ', '_').to_sym
+        @status = options[:status] || meta[0]
+        super(options)
       end
     end
 
@@ -108,6 +108,7 @@ module Pliny
       UnprocessableEntity          => [422, 'Unprocessable entity'],
       TooManyRequests              => [429, 'Too many requests'],
       InternalServerError          => [500, 'Internal server error'],
+      HTTPStatusError              => [500, 'Unspecified'],
       NotImplemented               => [501, 'Not implemented'],
       BadGateway                   => [502, 'Bad gateway'],
       ServiceUnavailable           => [503, 'Service unavailable'],
--- a/lib/pliny/errors.rb
+++ b/lib/pliny/errors.rb
@@ -1,16 +1,17 @@
 module Pliny
   module Errors
     class Error < StandardError
-      attr_accessor :id
+      attr_accessor :id, :metadata
 
       def self.render(error)
         headers = { "Content-Type" => "application/json; charset=utf-8" }
-        data = { id: error.id, message: error.message }
+        data = { id: error.id, message: error.message }.merge(error.metadata)
         [error.status, headers, [MultiJson.encode(data)]]
       end
 
       def initialize(options={})
         @id = options[:id]
+        @metadata = options[:metadata] || {}
         super(options[:message])
       end
     end
--- a/spec/errors_spec.rb
+++ b/spec/errors_spec.rb
@@ -1,24 +1,28 @@
 require "spec_helper"
 
 describe Pliny::Errors do
-  it "includes a general error that requires an identifier" do
-    e = Pliny::Errors::Error.new("General error.", :general_error)
-    assert_equal "General error.", e.message
-    assert_equal :general_error, e.id
+  describe Pliny::Errors::Error do
+    it "takes a message" do
+      e = Pliny::Errors::Error.new(message: "Fail.")
+      assert_equal "Fail.", e.message
+    end
+    it "takes an identifier" do
+      e = Pliny::Errors::Error.new(id: :fail)
+      assert_equal :fail, e.id
+    end
   end
 
-  it "includes an HTTP error that will take generic parameters" do
-    e = Pliny::Errors::HTTPStatusError.new(
-      "Custom HTTP error.", :custom_http_error, 499)
-    assert_equal "Custom HTTP error.", e.message
-    assert_equal :custom_http_error, e.id
-    assert_equal 499, e.status
-  end
+  describe Pliny::Errors::HTTPStatusError do
+    it "includes an HTTP error that will take generic parameters" do
+      e = Pliny::Errors::HTTPStatusError.new(status: 499)
+      assert_equal 499, e.status
+    end
 
-  it "includes pre-defined HTTP error templates" do
-    e = Pliny::Errors::NotFound.new
-    assert_equal "Not found.", e.message
-    assert_equal :not_found, e.id
-    assert_equal 404, e.status
+    it "includes pre-defined HTTP error templates" do
+      e = Pliny::Errors::NotFound.new
+      assert_equal "Not found.", e.message
+      assert_equal :not_found, e.id
+      assert_equal 404, e.status
+    end
   end
 end
--- a/spec/errors_spec.rb
+++ b/spec/errors_spec.rb
@@ -10,6 +10,12 @@ describe Pliny::Errors do
       e = Pliny::Errors::Error.new(id: :fail)
       assert_equal :fail, e.id
     end
+
+    it "takes metadata" do
+      meta = { resource: "artists" }
+      e = Pliny::Errors::Error.new(metadata: meta)
+      assert_equal meta, e.metadata
+    end
   end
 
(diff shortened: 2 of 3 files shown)


lib/pliny/errors.rb
```diff
@@ -108,6 +109,7 @@ class GatewayTimeout < HTTPStatusError; end               # 504
       UnprocessableEntity          => [422, 'Unprocessable entity'],
       TooManyRequests              => [429, 'Too many requests'],
       InternalServerError          => [500, 'Internal server error'],
+      HTTPStatusError              => [500, 'Unspecified'],
```
brandur: @pedro The original idea here was that `HTTPStatusError` would behave more like an abstract class which would never be instantiated directly; which is also why it didn't get an entry in this table. What do you think of changing this to just raising an `ArgumentError` if the instantiated error class wasn't found in `META`?

lib/pliny/errors.rb
```diff
@@ -1,29 +1,30 @@
...
     class HTTPStatusError < Error
       attr_accessor :status
 
-      def initialize(message = nil, id = nil, status = nil)
-        meta    = Pliny::Errors::META[self.class]
-        message = message || meta[1] + "."
-        id      = id || meta[1].downcase.tr(' ', '_').to_sym
-        @status = status || meta[0]
-        super(message, id)
+      def initialize(options={})
```
brandur: What do you think about changing this signature to:

``` ruby
def initialize(message, id: nil, status: nil)
  ...
```

Which the rough justification:

1. Having exceptions take a message as their first argument is a fairly widespread convention.
2. Optional arguments are safer than options hashes.

spec/errors_spec.rb
```diff
@@ -1,24 +1,34 @@
...
-  it "includes a general error that requires an identifier" do
-    e = Pliny::Errors::Error.new("General error.", :general_error)
-    assert_equal "General error.", e.message
-    assert_equal :general_error, e.id
-  end
+  describe Pliny::Errors::Error do
+    it "takes a message" do
+      e = Pliny::Errors::Error.new(message: "Fail.")
+      assert_equal "Fail.", e.message
+    end
```
brandur: Maybe space out the whitespace here.

spec/errors_spec.rb
```diff
@@ -1,24 +1,34 @@
 require "spec_helper"
 
 describe Pliny::Errors do
-  it "includes a general error that requires an identifier" do
-    e = Pliny::Errors::Error.new("General error.", :general_error)
-    assert_equal "General error.", e.message
-    assert_equal :general_error, e.id
-  end
+  describe Pliny::Errors::Error do
```
brandur: Nice job adding a little more test coverage here!

brandur:
@pedro Nice job. Left a few minor comments, but +1 from me!

lib/pliny/errors.rb
pedro: ah, makes sense! changing it.

gudmundur:
@pedro I'd love to see this merged. Is there anything that needs to be fixed or can it be merged?

pedro:
@brandur @gudmundur hey folks, good time to follow up on this – I just had to refactor how we handle errors at Lugg and ended up with a mix of Pliny and new stuff we're super happy about.

It's not a huge departure from this. This basically highlights the issue:

> Having exceptions take a message as their first argument is a fairly widespread convention.

Ruby exceptions are meant to have a message, but we all know API errors really just need an id. They are easier to compare in tests, and they won't pollute code with copy that we all know should live under `config/locales.yml`.

But no one said that the Ruby exception message is what we render to end-users. "invalid_json" works perfectly fine as the internal exception message, and it also happens to be a good error identifier for Pliny APIs, not to mention a good key to lookup the corresponding entry in the locales file.

So here's the proposal: Pliny ships with i18n. All existing user error messages are set there (like "Unauthorized" and "Unprocessable entity").  Our base error class only requires an id, which is used to fetch these messages and save them as the `error.user_message`. The id is also stored internally as the exception message so you can raise them using the same interface, like `raise Pliny::Errors::BadRequest("invalid_json")`.

You can then test for these exceptions using standard matchers, like:

```ruby
expect { run }.to raise_error(Pliny::Errors::BadRequest, "invalid_json")
```

Finally, while in there I'd also like to remove all exception classes outside 4xx and 5xx. Because if you want to render a 30x, for instance, you should rely on Sinatra's helpers and not on a Pliny exception.

Thoughts?

mattreduce:
@pedro :100: :+1:

brandur:
> Thoughts?

I like it (both taking the messages out with default localization and removing non-4/5xx series exceptions). +1.

dmathieu:
I really like this idea as well.

pedro:
you folks rock, thanks for the feedback! updated to match then :dancer:

dmathieu:
This is looking nice. The only comment I have is from a non-native english perspective. If I were to wish to translate error messages, I wouldn't be able to do the following:

```
error = Pliny::Errors::BadRequest.new
I18n.locale = :fr
error.user_message #=> will show the error in english
```

It's probably an edge case though.

lib/pliny/errors.rb
```diff
@@ -1,117 +1,61 @@
...
-    class RequestedRangeNotSatisfiable < HTTPStatusError; end # 416
-    class ExpectationFailed < HTTPStatusError; end            # 417
-    class UnprocessableEntity < HTTPStatusError; end          # 422
-    class TooManyRequests < HTTPStatusError; end              # 429
-    class InternalServerError < HTTPStatusError; end          # 500
-    class NotImplemented < HTTPStatusError; end               # 501
-    class BadGateway < HTTPStatusError; end                   # 502
-    class ServiceUnavailable < HTTPStatusError; end           # 503
-    class GatewayTimeout < HTTPStatusError; end               # 504
+    def self.MakeError(status, id)
```
brandur: Shouldn't this method name be snakecased?

brandur:
@dmathieu Should be easy enough to just move the translation over to be execute on access of `user_message`, but I guess you do introduce a slight risk of making rendering an error multiple times more expensive. IMO, either deficiency in likely to cause too much of a problem in practice, so will leave it up to Pedro :)

@pedro Left a minor comment above, but looks good to me! +1.

dmathieu:
My thinking was more to actually lazily fetch this data (and maybe memoize it to avoid expensive i18n lookup calls).

```
def user_message
  @user_message ||= I18n.t("errors.#{id}")
end
```

brandur:
> My thinking was more to actually lazily fetch this data (and maybe memoize it to avoid expensive i18n lookup calls).

Ah! I guess the only thing there is that with that implementation you could produce another failure scenario that's not too different from the one you have above:

``` ruby
error = Pliny::Errors::BadRequest.new
error.user_message #=> memoized
I18n.locale = :fr
error.user_message #=> will show the error in english
```

Although admittedly, also probably unlikely in practice.

lib/pliny/errors.rb
pedro: oh god, yes. I think I saw this class-looking method used before for a similar purpose but looking now it's super confusing. Fixed!

pedro:
good point. moved to a method not memoized for now, don't think we should worry too much about optimizing exceptional paths.

dmathieu:
:+1:

jellybob:
If that's actually an issue in practice could you memoize keyed against the current locale, something like `#{I18n.locale}_#{key}` which then be invalidated by changing the locale?

lib/pliny/errors.rb
```diff
@@ -1,117 +1,64 @@
...
-    class NoContent < HTTPStatusError; end                    # 204
-    class ResetContent < HTTPStatusError; end                 # 205
-    class PartialContent < HTTPStatusError; end               # 206
-    class MultipleChoices < HTTPStatusError; end              # 300
-    class MovedPermanently < HTTPStatusError; end             # 301
-    class Found < HTTPStatusError; end                        # 302
-    class SeeOther < HTTPStatusError; end                     # 303
-    class NotModified < HTTPStatusError; end                  # 304
-    class UseProxy < HTTPStatusError; end                     # 305
-    class TemporaryRedirect < HTTPStatusError; end            # 307
```
gudmundur: I know that removing all the non 4xx-5xx error classes is going to cause things to break in a few of my apps. To make matters a little worse, I suspect a few of them might be used for control flow.

That said, I do think this is a good change and what I've been doing is BAD (tm).

lib/pliny/errors.rb
```diff
@@ -1,117 +1,64 @@
...
-      attr_accessor :status
+      attr_accessor :id, :status, :metadata
 
-      def initialize(message = nil, id = nil, status = nil)
-        meta    = Pliny::Errors::META[self.class]
-        message = message || meta[1] + "."
-        id      = id || meta[1].downcase.tr(' ', '_').to_sym
-        @status = status || meta[0]
-        super(message, id)
+      def initialize(id=nil, metadata: {})
```
gudmundur: This is a breaking API change which will cause me some pain in a few places. Mostly where we are raising errors with custom messages, for example:

```ruby
fail Pliny::Errors::BadRequest, MY_CUSTOM_MESSAGE
```

Again, I agree that using an error `id` and `i18n` to sort this out is a better solution long term.

pedro:
@gudmundur nod on the breaking changes – this would obviously require a major version bump normally, but seeing that we're still on the 0.x series I'm not sure there's much else we can do. Thoughts?

brandur:
@pedro It's kind of a departure from the proposal above, should we allow an override message to be given for backward compatibility (and make that route deprecated, or at least unusual)? I guess the disadvantage is developers not knowing that they usually shouldn't provide it, but that's not a huge downside.

Definitely like the original proposal as is, but given that Pliny's seeing a reasonable amount of use now, erring on the side of conservative interface changes is probably a good idea.

PR #216, opened 2015-09-30
Fix rake schema

hayduke19us:
I noticed that the `rake schema` task was throwing a NoMethodError.
In *lib/tasks/schema.rb* the Pliny::Commands::Generator::Schema.new(nil).rebuild.

The error was occurring because the private method in the Base class `#normalize_name` was
trying to `#underscore` nil.

I added a ternary in the initialization for the name attribute. 

Basically Don't normalize if name is nil!


brandur:
Thanks @hayduke19us! Great tests as well.

PR #218, opened 2015-10-08
Convert Instruments from extension to middleware

gudmundur:
Changes the `Instruments` extention to a middleware, as I've been running into issues with the extension logging `status=500` for raised Pliny errors. Discussed this with @pedro and it seem the issue arises from the instrumentation being done from an extension.

This changes a couple of things from before:

1. When upgrading pliny, users will have to add the middleware and remove the extension.
2. `ROUTE_SIGNATURE` is not available in the middleware.


pedro:
Yeeees! wasted quite some time before debugging a 401 that was instrumented as a 200 because it actually came from the `Rack::Protection` middleware...

I really love the route signature though. The good news is that [starting in Sinatra 1.4.6 `sinatra.route` should be available in the Rack environment](https://github.com/sinatra/sinatra/commit/27434c13d4c8050bc80ef748f8657ac6c6545c8c). Will open another pull to get that back in.

pedro:
See #219

gudmundur:
@pedro Nice! :D Thanks for this.

ys:
woot thanks for this. YES debugging logs with that kind of errors happened
to us too:D

PR #220, opened 2015-10-10
Add GET /health and /health/db endpoints

gudmundur:
For every pliny project I've done, I've added a `GET /health` endpoint to hook up to Pingdom. To reduce the use of my mad copy pasting skills, I'm upstreaming it.

```
$ curl -i :5000/health
HTTP/1.1 200 OK
Content-Type: application/json;charset=utf-8
X-Content-Type-Options: nosniff
Vary: Accept-Encoding
Request-Id: 1620b575-4033-40a7-a8e6-17a1a4c94523
Content-Length: 2

{}
```

Also adds `GET /health/db` to test the Sequel database connection via `DB.test_connection`

### DIFF
--- a/lib/template/spec/endpoints/health_spec.rb
+++ b/lib/template/spec/endpoints/health_spec.rb
@@ -1,0 +1,16 @@
+require "spec_helper"
+
+describe Endpoints::Health do
+  include Rack::Test::Methods
+
+  def app
+    Endpoints::Health
+  end
+
+  describe 'GET /health' do
+    it 'returns a 200' do
+      get '/health'
+      assert 200, last_response.status
+    end
+  end
+end
--- a/lib/template/spec/endpoints/health_spec.rb
+++ b/lib/template/spec/endpoints/health_spec.rb
@@ -10,7 +10,10 @@ describe Endpoints::Health do
   describe 'GET /health' do
     it 'returns a 200' do
       get '/health'
-      assert 200, last_response.status
+      assert_equal(200, last_response.status)
+      assert_equal('application/json;charset=utf-8', last_response.headers['Content-Type'])
+      assert_equal(2, last_response.headers['Content-Length'].to_i)
+      assert_equal({}, MultiJson.decode(last_response.body))
     end
   end
 end
--- a/lib/template/spec/endpoints/health_spec.rb
+++ b/lib/template/spec/endpoints/health_spec.rb
@@ -1,6 +1,6 @@
 require "spec_helper"
 
-describe Endpoints::Health do
+RSpec.describe Endpoints::Health do
   include Rack::Test::Methods
 
   def app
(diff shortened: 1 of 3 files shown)


ys:
It is an API, I would still use a json response here.
But otherwise I kind of like to see that here.

It is one of our good practice. If it doesn't do it here, maybe we can add it as a generator in a side project.

gudmundur:
@ys Excellent point. Updated the PR to respond with JSON.

ys:
:heart:

dmathieu:
Since pliny will setup sequel by default, I think an health endpoint should make sure the database is available.
Something like `$db.test_connection`.

pedro:
AWESOME! And +1, can we has the connection test? :dancer:

gudmundur:
@dmathieu @pedro I've added `GET /health/db`. It wasn't as trivial as I hoped for since Pliny doesn't assign a global `$db` or `DB`. Please take a look and let me know if you've got a better way to do this.

pedro:
aah right on, in a few Pliny projects now I changed the initializer to save that db instance. Opened a pull to make that official #223

gudmundur:
@pedro Thank you! I'll get this rebased when I get back to coding.

gudmundur:
@pedro I've updated this PR to use the `DB` constant from #223. When that PR is merged, I'll rebase this one to clean up everything and resolve the conflict.

lib/template/spec/endpoints/health_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      allow(DB).to receive(:test_connection).and_return(false)
+
+      assert_raises Pliny::Errors::ServiceUnavailable do
+        get '/health/db'
+      end
+    end
+
+    it 'returns a 200' do
+      connection = double('connection', test_connection: true)
+      allow(Sequel::DATABASES).to receive(:first).and_return(connection)
```
soulim: Would it make sense to stub `DB.test_connection` as it's already done [here][1] and [here][2]?

```ruby
allow(DB).to receive(:test_connection).and_return(true)
```

Btw, I :heart: the idea of `/health` endpoint! Thank you @gudmundur!

[1]: https://github.com/gudmundur/pliny/blob/e75a74bf923b2ec229e91075e56ef03b73ea113c/lib/template/spec/endpoints/health_spec.rb#L30
[2]: https://github.com/gudmundur/pliny/blob/e75a74bf923b2ec229e91075e56ef03b73ea113c/lib/template/spec/endpoints/health_spec.rb#L38

gudmundur: Good catch. Missed this one while doing the `DB` const update.

gudmundur:
@brandur @pedro Rebased and ready to go.

ys:
<3

dmathieu:
:blue_heart:

soulim:
:heart:

brandur:
@gudmundur Thank-you @gudmundur!

PR #222, opened 2015-10-13
Keep `spec_helper.rb` lean

soulim:
* DatabaseCleaner configuration moved into `spec_support/database_cleaner.rb`
* Database seeding moved into `spec_support/seed_database.rb`

This is follow-up for https://github.com/interagent/pliny/pull/206.

### DIFF
--- a/lib/template/spec/spec_helper.rb
+++ b/lib/template/spec/spec_helper.rb
@@ -18,23 +18,6 @@ require_relative "../lib/initializer"
 Pliny::Utils.require_glob("#{Config.root}/spec/spec_support/**/*.rb")
 
 RSpec.configure do |config|
-  config.before :suite do
-    DatabaseCleaner.clean_with(:truncation)
-    DatabaseCleaner.strategy = :transaction
-  end
-
-  config.before :all do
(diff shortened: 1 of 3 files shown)


brandur:
Hey @soulim, thanks a lot for the pull here!

While I agree that in general that separation of concerns is nice, I think there's something to be said here for keeping the important paths that happen as part of a standard test run very explicit and an in-your-face part of `spec_helper`. That way, if you're ever running into a test-related problem, it's pretty easy to go in there an figure out exactly how all the plumbing works. It also allows us perfect control over the order that each test-related task runs in.

Simplecov is a bit of a special case because its whole operation is so tangential to the important spec suites.

Anyway, let me know if you (or anyone else on here) disagrees, and we can definitely revisit. Thanks again for the patch!

soulim:
Thank you for the explanation @brandur!

I guess it's just a matter of internal conventions of each team or developer. Sometimes `spec_helper.rb` tends to grow because of many third-party tools configured there. It doesn't help with readability :smile: 
Personally I prefer to use `spec_helper.rb` just for RSpec configuration, and each third-party config goes into separate file that is required in `spec_helper.rb`.

PR #223, opened 2015-10-19
DB constant

pedro:
Make a `DB` constant available to the app and seeds, useful for non-model specific queries (eg: `DB.transaction`, `DB.test_connection`, etc).

Checked it works with Puma workers.

### DIFF
--- a/lib/pliny/tasks/db.rake
+++ b/lib/pliny/tasks/db.rake
@@ -34,7 +34,7 @@ begin
     task :seed do
       if File.exist?('./db/seeds.rb')
         database_urls.each do |database_url|
-          Sequel.connect(database_url)
+          DB = Sequel.connect(database_url)
           load 'db/seeds.rb'
         end
         disconnect
--- a/lib/pliny/tasks/db.rake
+++ b/lib/pliny/tasks/db.rake
@@ -34,7 +34,9 @@ begin
     task :seed do
       if File.exist?('./db/seeds.rb')
         database_urls.each do |database_url|
-          DB = Sequel.connect(database_url)
+          # make a DB instance available to the seeds file
+          self.class.send(:remove_const, 'DB') if self.class.const_defined?('DB')
+          self.class.const_set('DB', Sequel.connect(database_url))
           load 'db/seeds.rb'
         end
         disconnect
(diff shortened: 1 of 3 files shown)


lib/pliny/tasks/db.rake
```diff
@@ -34,7 +34,7 @@ begin
     task :seed do
       if File.exist?('./db/seeds.rb')
         database_urls.each do |database_url|
-          Sequel.connect(database_url)
+          DB = Sequel.connect(database_url)
```
bjeanes: this is iterating over `database_urls`, which means it'll keep setting the same constant multiple times when more than 1 urls are present. Is that intended?

pedro: ooh true, this causes a warning to be displayed.

It is intended though, I wanted `db/seeds.rb` to have an instance of `DB` just like the rest of the application. Thoughts?

bjeanes: But if there's multiple URLs, the `DB` might be pointing to an arbitrary database (i.e. it's order dependent). Does that really mean it does the correct thing? I don't know enough about the scenarios in which there are multiple URLs to provide a recommendation. If you know for sure that it's always the first one, then making `each` a `map` that returns the connection and saving the first one would be better, for example.

gudmundur: I fail to see the negative consequences of this, other than the constant being set after the `each` is done. My understanding here is that for each database, it'll set the `DB` constant before loading the seeds. Am I missing something else?

pedro: Hey! Getting back to this.

So I just silenced the const warning. Not pretty, but it's code in Pliny exactly to make sure the Pliny app's code is not ugly :}

pedro:
CI error is unrelated, it seems like json_schema is not compatible with 1.9.3 anymore.

@brandur going to drop 1.9 compat ok?

brandur:
> @brandur going to drop 1.9 compat ok?

Sounds good to me!

pedro:
Done, just removed from tests.
The template app hasn't been on 1.9 for way too long so not too concerned.

gudmundur:
@pedro :boom: Nice! Any chance of a tagged release for this?

PR #225, opened 2015-10-23
set search_path on schema dumps

### LINKED ISSUE #224: Cannot `rake db:schema:load` the result of `rake db:schema:dump`
I'm trying to synchronize my local dev environment with a pre-existing app.

Given the **schema.sql** from `rake db:schema:dump` on the existing Pliny app, I do the following:
1. `rake db:create`
2. `rake db:schema:load`

  ```
rake aborted!
Sequel::DatabaseError: PG::UndefinedTable: ERROR:  relation "schema_migrations" does not exist
LINE 912: INSERT INTO "schema_migrations" ("filename") VALUES ('144485...
                      ^
  ```

Looking through the source for these rake tasks, it looks like [the `dump` task outputs these](https://github.com/interagent/pliny/blob/master/lib/pliny/tasks/db.rake#L103) ` INSERT INTO "schema_migrations" …` commands, but [the `load` task does not](https://github.com/interagent/pliny/blob/master/lib/pliny/tasks/db.rake#L75) create the `"schema_migrations"` table, so it fails with this error.

This **schema.sql** contains structure for the Heroku Connect `salesforce` schema too, so I cannot fall back on the app's migrations for this.

What is the right way to use `rake db:schema:dump` and `rake db:schema:load`? Would a PR to make the `load` task create the `"schema_migrations"` table be acceptable?

pedro:
This ensures that the generated schema can be restored even on
databases with different schemas, where the original pg_dump might
leave the search_path on a place where schema_migrations is not
visible.

Fix #224

### DIFF
--- a/lib/pliny/tasks/db.rake
+++ b/lib/pliny/tasks/db.rake
@@ -100,12 +100,11 @@ begin
 
         # add migrations used to compose this schema
         db = Sequel.connect(database_urls.first)
-
-        # set the search_path so migrations are added in the right schema
-        search_path = db.dataset.with_sql("SHOW search_path").single_value
-        schema << "SET search_path = #{search_path};\n\n"
-
         if db.table_exists?(:schema_migrations)
+          # set the search_path so migrations are added in the right schema
+          search_path = db.dataset.with_sql("SHOW search_path").single_value
+          schema << "SET search_path = #{search_path};\n\n"
+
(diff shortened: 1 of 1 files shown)


pedro:
Merging to fix an issue we're having with Pliny + Heroku Connect, but let me know if you have any feedback here.

PR #226, opened 2015-10-25
Enable RSpec zero monkey patching mode

soulim:
RSpec 3 configuration has [`disable_monkey_patching!`][1] option that removes monkey patching of the methods like `describe` and `shared_examples_for` onto `main` and `Module`. This should be default in RSpec 4.

This change will require update for existing specs on apps, but [transpec][2] gem might help with that.

I would be happy to know your thought on this one. Thank you for doing awesome job! :smile: 

[1]: http://www.rubydoc.info/gems/rspec-core/RSpec/Core/Configuration#disable_monkey_patching%21-instance_method
[2]: https://github.com/yujinakayama/transpec

### DIFF
--- a/lib/pliny/templates/endpoint_acceptance_test.erb
+++ b/lib/pliny/templates/endpoint_acceptance_test.erb
@@ -1,6 +1,6 @@
 require "spec_helper"
 
-describe Endpoints::<%= plural_class_name %> do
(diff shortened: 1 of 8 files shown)


brandur:
I like it! Most existing projects should already have a well-defined `spec_helper`, so I don't think this will break anybody. Thanks!

PR #227, opened 2015-10-25
Fix pliny-new command banner message

soulim:
If I'm not mistaken `pliny-new` accepts no options except an app name. It was a bit confusing when I've tried to find possible options :smile: 

Also I took a chance and messaged format of the banner message, so it matches the same style `pliny-generate` has. Consistency always matters :smiley:


brandur:
Good call! Thanks again @soulim!

> Consistency always matters :smiley:

So true!

PR #229, opened 2015-11-11
Re-raising the error from the root cause rather than raising a new error

reidmix:
Otherwise the raise will show the backtrace from the `log`/`log_to_stream` method.


brandur:
Thanks Reid! We appreciate the added tests as well :)

brandur:
Released as 0.13.1.

PR #232, opened 2015-12-11
Make rack-timeout a dependency of the template app

pedro:
Just makes more sense, pliny doesnt need it internally


dmathieu:
Meaning we can also remove (bad idea) or use another version. Good idea. :+1:

brandur:
Definitely a good call. +1.

pedro:
Cool, released in 0.14. /cc @gudmundur I think you were waiting for the `DB` changes so they're out too! next up we need to get you to release :}

gudmundur:
Thanks @pedro. I was planning on doing a release with the `DB` changes, but no need now! :dancer:

PR #238, opened 2016-01-15
Remove rack_env and make pliny_env mandatory

### LINKED ISSUE #237: The PLINY_ENV default bubbles to RACK_ENV
If `PLINY_ENV` is left unset on staging or production environments, it causes Puma to override `RACK_ENV` to `development`, [see here](https://github.com/puma/puma/blob/e81a1326ea7a791d7729b7674b5bebf9409657aa/lib/rack/handler/puma.rb#L21-L23).

What is the motivation behind `PLINY_ENV` and how should it differ from `RACK_ENV`? Should it its default be equal to `ENV['RACK_ENV']`?

dmathieu:
That's my way of saying I agree with [brandur's comments](https://github.com/interagent/pliny/pull/138#discussion_r49124034).

Closes #237

### DIFF
--- a/lib/template/config/config.rb
+++ b/lib/template/config/config.rb
@@ -11,6 +11,7 @@ module Config
 
   # Mandatory -- exception is raised for these variables when missing.
   mandatory :database_url, string
+  mandatory :pliny_env,    string
 
   # Optional -- value is returned or `nil` if it wasn't present.
   optional :placeholder,         string
@@ -22,13 +23,11 @@ module Config
   override :db_pool,          5,    int
   override :deployment,       'production', string
   override :force_ssl,        true,  bool
-  override :pliny_env,        'development', string
   override :port,             5000, int
   override :pretty_json,      false, bool
   override :puma_max_threads, 16,   int
   override :puma_min_threads, 1,    int
   override :puma_workers,     3,    int
-  override :rack_env,         'development', string
   override :raise_errors,     false,         bool
   override :root,             File.expand_path("../../", __FILE__), string
   override :timeout,          10,    int


brandur:
+1 from my point of view! Thanks for putting this together Damien!

/cc @gudmundur Does this sound reasonable to you?

gudmundur:
Thanks Damien. Since `pliny_env` is mandatory now, shouldn't we add default values to `.env.sample` and `.env.test`?

dmathieu:
They're [already](https://github.com/interagent/pliny/blob/master/lib/template/.env.sample#L2) [there](https://github.com/interagent/pliny/blob/master/lib/template/.env.test#L2) :)

gudmundur:
@dmathieu doh! Thanks for pointing that out.

When I think about how it is to deploy a pliny app for the first time to Heroku, there are a few steps involved. Creating the app, provisioning the database, pushing the code and running the migrations. By making `PLINY_ENV` mandatory, we'd be adding an additional step that doesn't feel quite intuitive to me since there are defaults for dev and tests. What do you think about leaving `PLINY_ENV` as an `override` but changing the default to `production`?

jellybob:
It might be more work than people are up for, but can buildpacks set environment variables, and if so could we provide one for Pliny that sets sane defaults? Even better if it can be autodetected, but that would definitely be a nice to have.

dmathieu:
@jellybob a custom buildpack would probably not be a good idea, as that would add quite a lot of additional maintenance work compared to the advantages. Also, pliny is not an heroku-only framework. You might not want to host your app there.

I'm not convinced of defaulting to production when no environment is set. No other library around does the same AFAIK. It'd be interesting to know why.
I see we don't generate any app.json file. How about adding one with that config?

gudmundur:
@dmathieu Having an `app.json` is an excellent idea. There are currently two deficiencies in App Setups that make it less useful and confusing, IMO.

1) They're not picked up on `git push heroku master`.
2) They prime a Git repository based on the tarball fetched from GitHub. The subsequent push has to be a force push.

What is your thinking by not changing the default to `production`?

jellybob:
My concern with changing the default to production would be apps that are enabling things like sending emails to actual customers based off `PLINY_ENV`. While things like which SMTP server to use *should* be configured via environment variables I have little doubt there will be people who don't, and the first they know of it is when they accidentally email their entire customer base thinking they were doing a test run.

dmathieu:
Agreed that defaulting to production does cause issue if you also have configuration within your code and you run the app in that mode mistakenly. I don't think this is a good reason though.
I don't think we should go one way because it will prevent people using non-recommended practices from making mistakes.

My concern is more one of least surprise. If you do end up on the production environment locally, and start seeing public error pages instead of actual exceptions, it could take several hours for anyone to understand what's going on.
Erroring out right away if you don't set the environment makes it much more explicit what the actual reason for the problem is.

jellybob:
+1 on just erroring out. I hadn't quite made the leap that marking the variable mandatory would prevent the application starting up with the wrong config.

dmathieu:
It's not preventing it from starting up with a "wrong config" (dev instead of production). It's preventing it from starting with a missing config (PLINY_ENV not being set).

gudmundur:
@jellybob @dmathieu Thanks for taking the time to explain your thinking on this.

I think there's a misunderstanding between us in what `PLINY_ENV` means. Right now, on `master`, the effect of it, is overriding and masking `RACK_ENV`. With this change we're making it mandatory, forcing users to set it on their staging/production environment, and it is still masking `RACK_ENV`.

What I'm understanding from your comments is that you wish to use `PLINY_ENV` for various purposes, and to me that sounds like we are conflating a few different concepts. For example, which stage in a pipeline an app is running, already has a config var in Pliny called `deployment`. Similarly, a similar discussion, which Rollbar environment is in use, I think that should be a separate config variable.

How about we remove `pliny_env`, bring back `rack_env` and make that mandatory? It'll do all the right things in dev, staging and production (on Heroku).

dmathieu:
Discussed this directly with @gudmundur. Closing in favor of #244.

PR #240, opened 2016-01-16
Add app_name config variable and log it by default

gudmundur:
When creating new Pliny apps, one of the first things I do is to add `app_name` as a config var, and log it out by default.

When calling `Pliny.log(hello_world: true)` it'll yield the log message

```
app=my-fancy-app hello_world
```

This change makes it easy to search for log messages from a single app when multiple apps are being pooled into the same Splunk instance.

**Note** This is not the same as the Heroku app name, that config var will be set automatically in `HEROKU_APP_NAME` if [dyno metadata is enabled](https://devcenter.heroku.com/articles/dyno-metadata).

### DIFF
--- a/lib/template/config/initializers/log.rb
+++ b/lib/template/config/initializers/log.rb
@@ -1,3 +1,2 @@
-Pliny.default_context = {
-  app: Config.app_name
-}
+Pliny.default_context = {}
+Pliny.default_context[:app] = Config.app_name if Config.app_name
--- a/lib/template/config/initializers/log.rb
+++ b/lib/template/config/initializers/log.rb
@@ -1,1 +1,3 @@
-Pliny.default_context = {}
+Pliny.default_context = {
+  app: Config.app_name
+}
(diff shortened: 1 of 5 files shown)


lib/template/config/initializers/log.rb
```diff
@@ -1 +1,3 @@
-Pliny.default_context = {}
+Pliny.default_context = {
+  app: Config.app_name
```
pedro: I think this is going to start logging `app=` when the config is not set, right?
Maybe only add the key "app" if the config is set?

pedro:
:+1:!

dmathieu:
It seems weird to set the default app name to `pliny`. What about setting it to the name of the folder the pliny app is created on?

gudmundur:
@dmathieu The `pliny-new` generator will override the default values in `.env.sample` and `.env.test` to the name you pass in.

dmathieu:
Oh right, the gsub my bad. LGTM (I agree with Pedro's inline comment though).

dmathieu:
LGTM, though if we need to add other dynamic changes to `.env` files, we should probably parse them with ERB to avoid doing many gsub.

gudmundur:
@dmathieu I agree.

PR #241, opened 2016-01-16
Upgrade to Ruby 2.3.0 and upgrade pinned gems

gudmundur:
Upgrades the template to the latest version of Ruby and bumps the version numbers for Puma and Sequel.


pedro:
To start a tangential thread here: I think in the past we've been bumping Pliny dependencies pretty early. As a user, I'd love to know that Pliny gives me versions that have been validated in production to some extent.

It's pretty ambitious, but maybe there are small steps we can take towards this direction. For instance:

- If you tell me the Heroku API (or another app of similar scale/usage) has been running a certain version I'd be totally comfortable to update myself. Knowing that not only it was in production but also that someone was on call looking after it is huge.
- We all have different standards for different dependencies; I wouldn't think twice before bumping a minor in Sequel, but going for Ruby 2.3 when it's the 1st of this series and only a few weeks old seems significantly riskier.
- Prereleases are a great way to give people things that haven't been battle tested, but we can't possibly expect people to test everything that we push out. [Rack does a great job in finding a balance here](https://rubygems.org/gems/rack/versions) – it has only a few pre-release versions, for the stuff they consider risky. Maybe after 1.0 (and assuming we have a big enough user base) we can aim for something similar.

brandur:
@pedro I can definitely see your point on this one, but it might be a mistake to err too far on the side of conservatism for a smallish project like this one; especially if we don't really have a sufficient story around how we'd actually go about adequately vetting new versions. A pre-release version might be a good idea, but we'd probably want to be sure that we actually have enough users willing to alpha test to make the effort worth it.

I know we used to make a big deal about Ruby version upgrades, but I think that most of the recent ones have been fairly happy (2.0 -> 2.1 -> 2.2). If I recall correctly, you did find some problems with increased memory use during one of those upgrades, but there's only so much you can really do there unless it's an actual bug — either you say on your old version forever or you just have to take the plunge and accept the consequences. We tried to do the former with AR 2.3 and it's unaddressed performance problems, but eventually we just went for the upgrade.

Not to say we should go directly to bleeding edge on everything, but there may be some balanced process that we can try here. e.g. Wait a month or two after any minor level or above upgrades and then bring them in unless some significant problems with those dependencies can be demonstrated to have been unaddressed. That should give enough time for most major bugs to have been weeded out in a lot of projects. Thoughts?

pedro:
I think we're actually on the same page: I'm not proposing that we get super conservative, but looking at this pull and a few others I do believe we're erring on the bleeding edge. Even if people took the time to try 2.3 during christmas and nye, it hasn't been a month since it's out.

So giving a month or two before bumping "delicate" dependencies like Ruby and Puma sounds like a reasonable start.

And in the long run we can do better, obviously. I'd still love to know that someone actually ran the thing in production before updating, this kind of curation is exactly where Pliny can bring even more value than Rails/Sinatra alone.

dmathieu:
The Heroku API is running on 2.3.0, as well as several other components such as kolkrabbi.

gudmundur:
@pedro @brandur Thanks for taking the time to raise these excellent points. I'm fine with waiting a month or two before having this merged.

geemus:
It's been a bit now and we have been running it in prod, so going to go ahead and bring this in. Thanks!

pedro:
+999! Thank you :}

PR #242, opened 2016-01-16
Add Serialize helpers

gudmundur:
My take on @ys's serialize helpers. The two of us discussed this out-of-band and this is my understanding of where we should head with this.

This PR supersedes #182.


pedro:
This looks great to me!

brandur:
+1 from me too!

You may want to consider sanitizing your commits a bit more to represent a slightly cleaner progression towards the implementation, but it's not a big deal.

gudmundur:
This was a nice exercise in rebasing to death. @brandur commits looking better?

gudmundur:
I'm thinking about adding a sanity check that if you call `serialize` without specifying a serializer, it'll blow up with a `500`, similar to what API does for missing permissions, etc. That, and some specs.

brandur:
> This was a nice exercise in rebasing to death. @brandur commits looking better?

Yep! Thanks for cleaning that up.

> I'm thinking about adding a sanity check that if you call serialize without specifying a serializer, it'll blow up with a 500, similar to what API does for missing permissions, etc. That, and some specs.

Nice idea! +1 if you want to try your hand at this. Please include a reasonable error message on failure :)

gudmundur:
@brandur there you go!

brandur:
@gudmundur Awesome! Looks good to me. Let's get this in :)

PR #243, opened 2016-01-16
Add deployment to default log context

gudmundur:
In conjunction with #240, the default log context will contain the app and the deployment, the benefit of doing this is that log entries look like this

```
app=my-pliny-app deployment=production
```

This PR will be rebased is and when #240 hits master.

### DIFF
--- a/lib/template/config/initializers/log.rb
+++ b/lib/template/config/initializers/log.rb
@@ -1,2 +1,3 @@
 Pliny.default_context = {}
 Pliny.default_context[:app] = Config.app_name if Config.app_name
+Pliny.default_context[:deployment] = Config.deployment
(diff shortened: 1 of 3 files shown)


pedro:
hawt :+1:

brandur:
+1. Feel free to go for the merge @gudmundur!

lib/template/config/initializers/log.rb
```diff
@@ -1,2 +1,3 @@
 Pliny.default_context = {}
 Pliny.default_context[:app] = Config.app_name if Config.app_name
+Pliny.default_context[:deployment] = Config.deployment
```
gudmundur: Intentionally setting this outside the `default_context` initialization to keep the key ordering of log entries as

```
app=my-app deployment=development
```

PR #244, opened 2016-01-22
Remove rack_env and set it to only always be development or deployment

dmathieu:
Rack doesn't accept "production" as a RACK_ENV variable. It requires
development or deployment.
http://unicorn.bogomips.org/unicorn_1.html#rack-environment

Setting it to another value means some middlewares will not be
automatically added:
https://github.com/rack/rack/blob/028438ffffd95ce1f6197d38c04fa5ea6a034a85/lib/rack/server.rb#L228

Not having those middlewares is known to cause latency with some
browsers.

### DIFF
--- a/lib/template/config/config.rb
+++ b/lib/template/config/config.rb
@@ -28,7 +28,6 @@ module Config
   override :puma_max_threads, 16,   int
   override :puma_min_threads, 1,    int
   override :puma_workers,     3,    int
-  override :rack_env,         'development', string
   override :raise_errors,     false,         bool
   override :root,             File.expand_path("../../", __FILE__), string
   override :timeout,          10,    int
--- a/lib/pliny/config_helpers.rb
+++ b/lib/pliny/config_helpers.rb
@@ -49,6 +49,14 @@ module Pliny
       end
     end
 
+    def rack_env
+      if pliny_env == "development" || pliny_env == "test"
+        "development"
+      else
+        "deployment"
+      end
+    end
+
     private
 
     def cast(value, method)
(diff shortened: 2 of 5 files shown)


lib/template/config/config.rb
```diff
@@ -1,5 +1,9 @@
 require "pliny/config_helpers"
 
+
+# Supress the "use RbConfig instead" warning.
+Object.send(:remove_const, :Config) if Object.const_defined?(:Config)
```
dmathieu: Moved this from `lib/pliny/config_helpers.rb` because we need to be able to require that helper without declaring the `Config` module.

brandur:
> Rack doesn't accept "production" as a RACK_ENV variable. It requires
> development or deployment.

Wow, that goes against everything I thought I knew about the variable! Honestly very surprising.

@dmathieu This patch looks like it's in great shape, so +1 from a technical perspective! I still don't have strong feelings about the use of these types of variables (mostly because I have yet to run into problems related to them), so I'll leave it up to you and @gudmundur to figure out how/when best to bring it in. Thanks!

dmathieu:
This was raised internally around a year ago by Keiko.
Another mention of this issue: http://www.hezmatt.org/~mpalmer/blog/2013/10/13/rack_env-its-not-for-you.html

lib/template/config/config.rb
```diff
@@ -1,5 +1,9 @@
 require "pliny/config_helpers"
 
+
+# Supress the "use RbConfig instead" warning.
```
gudmundur: Not yours, but would you mind fixing this typo: Supress => Suppress?

lib/template/config/config.rb
gudmundur: Would it be possible to leave this in the config helper, but add the module existence check to the conditional? I'd prefer to keep the templated config smooth.

gudmundur:
Awesome job @dmathieu! I've got one comment, but overall :+1:.

gudmundur:
Excellent work @dmathieu! Bringing this in. :dancer:

dmathieu:
:balloon:

PR #245, opened 2016-01-27
.env files parsed with ERB and add app.json

dmathieu:
This cleans up the gsubbing of .env files and introduces an app.json to the template.


gudmundur:
Nice work @dmathieu! Glad to see the `app.json` there.

PR #246, opened 2016-02-04
Add app.json postdeploy script to run migrations

gudmundur:
Adds `bundle exec rake db:schema:load db:migrate` as a `postdeploy` script when creating apps with Heroku App Setups. This will cause Heroku Buttons and review apps to get the database prepared while spinning up the app.


dmathieu:
:+1:

brandur:
+1.

PR #253, opened 2016-02-11
Upgrade to Ruby 2.3.0 and upgrade pinned gems - redux

gudmundur:
Reverts interagent/pliny#252. Brings back the Ruby 2.3.0 upgrade along with puma and sequel dependency bumps, #241.

The initial PR was reverted due to an issue in `sinatra-contrib`s Reloader, which caused a `Thread.exclusive is deprecated, use Mutex` and a stack trace for every incoming request while in development mode.

We should hold off with merging this until https://github.com/sinatra/sinatra-contrib/issues/183 has been resolved.


geemus:
Sounds good, thanks!

PR #254, opened 2016-02-11
Read ruby version from .ruby-version in Gemfile

tmaier:
Never again forget .ruby-version or Gemfile when updating the ruby version (example: 63cec6e1985fab54976b594c920131b56feffc8c)

### DIFF
--- a/lib/template/Gemfile
+++ b/lib/template/Gemfile
@@ -1,5 +1,5 @@
 source "https://rubygems.org"
-ruby "2.3.0"
+ruby IO.read(File.join(__dir__, ".ruby-version")).strip
 
 gem "multi_json"
 gem "oj"


dmathieu:
Why not just wait for bundler 2.0?
https://github.com/bundler/bundler/pull/4036/files

tmaier:
awesome. I was not aware they did this in bundler.

I don't know when bundler 2 will be released. If this is not foreseeable, I would implement this simplification till then.

brandur:
Nice. And wow, I'm looking forward to that Bundler upgrade ...

@gudmundur Any thoughts on this one?

soulim:
Interesting approach!

There's my two cents. In our projects we define (manually) Ruby version in Gemfile (only!), and then if you run `script/bootstrap`, it generates correct `.ruby-version` and `.ruby-gemset` files and makes sure require Ruby is available in the system. This way there's no need to additionally specify Ruby version for CI server and Heroku, because they rely on Gemfile.

```shell
  ruby_required=$(grep --regexp='^ruby' Gemfile | grep --regexp='\d\.\d\.\d*' --only-matching)
  ruby_gemset="${namespace}-${app}"

  echo "$ruby_required" > .ruby-version
  echo "$ruby_gemset" > .ruby-gemset

  ruby_installed=$(ruby --version | grep --regexp='\d\.\d\.\d*' --only-matching)
  if [ "$ruby_required" != "$ruby_installed" ]
  then
    echo "Ruby ${ruby_required} is required. Make sure it is installed"
    exit 1
  fi
```

So basically everything work other way round :smile:

gudmundur:
If Bundler tackles this soon, I'd be inclined to wait for that. @tmaier you alright with that path forward?

tmaier:
Fine for me. :)

PR #255, opened 2016-02-16
Fix code coverage by shuffling requirement order

fdr:
Previous to this patch, requiring "../lib/initializer" before the
spec_support helpers, simplecov is not present when requiring the
application code under test, and thus the generated coverage reports
are useless.


brandur:
Thanks for the fix!

PR #257, opened 2016-03-07
Remove the connection from `Sequel::DATABASES` when disconnecting in DbSupport

### LINKED ISSUE #247: Default migrate task does not clear active databases
The default migration task does not call Sequel::DATABASES.clear to remove the disconnected db instance from the global array. This later is causing an error downstream for me from DatabaseCleaner because it needs to know which connection to use if we have more than one active. It is a little funky that DatabaseCleaner is doing this. However I thought I'd file the issue for feedback since some of the other tasks are self contained in that they both disconnect the opened conn and clear it from the array.

Migration task in question:
https://github.com/interagent/pliny/blob/master/lib/pliny/tasks/db.rake#L15

Task disconnecting and clearing conn:
https://github.com/interagent/pliny/blob/master/lib/pliny/tasks/db.rake#L42

I can fix the migration task to match the others calling disconnect if it makes sense.

dmathieu:
Fix #247

### DIFF
--- a/lib/pliny/db_support.rb
+++ b/lib/pliny/db_support.rb
@@ -24,6 +24,7 @@ module Pliny
       instance = new(url, logger)
       yield instance
       instance.disconnect
+      Sequel::DATABASES.delete(instance)
     end
 
     attr_accessor :db


geemus:
Thanks!

PR #258, opened 2016-03-08
Rollbar environment

gudmundur:
See discussion here: https://github.com/interagent/pliny/pull/160.

@rwz what do you think?

### DIFF
--- a/lib/template/config/initializers/rollbar.rb
+++ b/lib/template/config/initializers/rollbar.rb
@@ -1,7 +1,7 @@
 Rollbar.configure do |config|
   config.enabled = ENV.has_key?('ROLLBAR_ACCESS_TOKEN')
   config.access_token = ENV["ROLLBAR_ACCESS_TOKEN"]
-  config.environment = ENV['ROLLBAR_ENVIRONMENT']
+  config.environment = ENV['ROLLBAR_ENV']
   config.use_sucker_punch
 end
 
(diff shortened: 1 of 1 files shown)


rwz:
:+1:

lib/template/config/initializers/rollbar.rb
```diff
@@ -1,6 +1,7 @@
 Rollbar.configure do |config|
   config.enabled = ENV.has_key?('ROLLBAR_ACCESS_TOKEN')
   config.access_token = ENV["ROLLBAR_ACCESS_TOKEN"]
+  config.environment = ENV['ROLLBAR_ENVIRONMENT']
```
soulim: Would it make sense to add `ROLLBAR_ENVIRONMENT` to [`pliny/lib/template/.env.sample.erb`](https://github.com/interagent/pliny/blob/fdf30c530367d07bae2064527019bfd38edbf7dc/lib/template/.env.sample.erb)?

```
ROLLBAR_ENVIRONMENT=development
```

soulim:
I know it's just cosmetics, but would it make sense to use `_ENV` suffix instead of `_ENVIRONMENT`? Then variable name would be consistent with `RACK_ENV`, `PLINY_ENV`, and etc.

lib/template/config/initializers/rollbar.rb
gudmundur: I intentionally left this out, as `ROLLBAR_ACCESS_TOKEN` isn't specified there. Which causes it to default to being disabled.

gudmundur:
@soulim Excellent point. I'll get it updated.

lib/template/config/initializers/rollbar.rb
soulim: Yeah, it makes sense. Thanks for explanation! :+1:

I wasn't sure how Rollbar gem acts when `environment` is `nil`, but seems like it [handles everything pretty well](https://github.com/rollbar/rollbar-gem/blob/d06b18d2a98d8effdde605355ebe14add7ae5d3d/lib/rollbar.rb#L368-L369) :smile:

PR #259, opened 2016-03-18
Logger proxy for Rollbar to use Pliny.log

gudmundur:
Proxies all Rollbar log entries through `Pliny.log`. The motivation for doing this is so that Rollbar messages and links will be included in log entries when searching, for example when filtering by `request_id`.

screen shot 2016-03-18 at 10 53 20

**Related**: https://github.com/interagent/pliny/pull/258


dmathieu:
:+1:

brandur:
Cool! +1.

PR #260, opened 2016-04-12
Handle api variants

dmathieu:
This adds support for Accept variants, as well as helpers to fetch the api version and variant in the controller.


lib/pliny/helpers/versioning.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    # A variant is a specially tagged version where new API endpoints can be
+    # experimented with. If a user passes a version into the `Accept` header
+    # like "3.env-vars", the `api_version` will be "3" and the `api_variant`
+    # will be "env-vars".
+    #
+    # By design, only a single variant can be specified at a time.
+    def api_variant
+      # set by Versioning middleware
+      env["HTTP_X_API_VARIANT"]
```
gudmundur: I know this is only used internally, but can you drop the `X_` infix? They've been "deprecated" in HTTP.

lib/pliny/middleware/versioning.rb
```diff
@@ -44,13 +44,14 @@ def detect_api_version(env)
...
       version ||= @default
-      set_api_version(env, version)
+      set_api_version(env, version, variant)
       @app.call(env)
     end
 
-    def set_api_version(env, version)
+    def set_api_version(env, version, variant)
       # API modules will look for the version in env
       env["HTTP_X_API_VERSION"] = version
```
gudmundur: I know changing the `X-` thing above might cause a breakage for this particular header. I wonder how we should migrate over, 1) just do it, 2) support both, 3) only stick to `X-`, 4) ???, 5) profit.

lib/pliny/helpers/versioning.rb
gudmundur: I just looked this up in API, and they are using `heroku.api_variant` for this. What do you think about using `pliny.api_variant`?

gudmundur:
Nice work @dmathieu. I was under the impression this was already supported in Pliny. Are there other aspects from API with regards to variants we should pull over? With regards to serializers, I think we're covered with the different `structure`s. The only other thing I can think of is `version` and `variant` helpers in mounting routes.

lib/pliny/helpers/versioning.rb
dmathieu: I thought of doing that, but it seems we wouldn't be able to set the header in the tests anymore.

lib/pliny/middleware/versioning.rb
dmathieu: I'm thinking we should just do it. The only way that header was officially exposed is through the `version` method in the router.

dmathieu:
We already have a [version helper](https://github.com/interagent/pliny/blob/master/lib/pliny/router.rb#L8) in the router.
There is no variant there, but do we need one? If you think so, I can add it.

lib/pliny/helpers/versioning.rb
dmathieu: Nevermind. b2a048a088b2963e8511923b31240e50e9af1b8c
I've used `api.version` instead of `pliny`, because pliny is the middleware, not the app name itself.

brandur: +1 on dropping `X_` and even `HTTP_`! These aren't really headers, so we should try to remove that conflation.

brandur:
I'll leave leadership to @gudmundur here, but I just wanted to drop in and say this looks good to me :) Thanks for the tests!

gudmundur:
@dmathieu Yeah, I think it makes sense to add one for the variant since we have one for the version. Also, I noticed `HTTP_X_API_VERSION` in the router version helper. What are your thoughts on `pliny.version` vs. `api.version`? IMO, we should be more specific.

dmathieu:
The latest commit in this PR removes all `HTTP_X_API_VERSION`, and replaces them with `api.version` (and does the same for variant). That includes the router version helper.

My opinion is that using `pliny` for this doesn't make much sense.
Yes, pliny sets it. But that's not the name of my app either.

gudmundur:
@dmathieu was thinking precisely this while biking just now. `api.*` makes perfect sense. I'll give this a better read in the next couple of days.

gudmundur:
@dmathieu this is looking great btw. If you don't mind doing a `variant` helper, I'd really appreciate that.

dmathieu:
Yes. I've just added that.

gudmundur:
What do you think of adding the `api_version` and `api_variant` to the log context? API currently just logs the version number, but I'm currently wishing it had variant as well. :)

dmathieu:
Thanks for the suggestion. I've added "api_version into the log context, as a merge of version and variant.

gudmundur:
Oh, one detail that we're missing here. The `rake schema` command is not handling schemas afaict.

lib/pliny/commands/generator/schema.rb
```diff
@@ -38,34 +46,38 @@ def warn_legacy
...
+          "./docs/schema"
+        else
+          "./schema"
+        end
+      end
+
       def schema_json_path
         if legacy?
-          "./docs/schema.json"
+          "#{schema_path}.json"
```
gudmundur: Shouldn't this be `"#{schema_path}/schema.json"`? Similar goes for the `meta.json`.

lib/pliny/commands/generator/schema.rb
```diff
@@ -38,34 +46,38 @@ def warn_legacy
...
+      def schema_path
+        if legacy?
+          "./docs/schema"
+        else
+          "./schema"
+        end
+      end
+
       def schema_json_path
         if legacy?
```
gudmundur: Actually, can't we drop this conditional and rely on `schema_path` providing the right thing?

lib/pliny/middleware/versioning.rb
```diff
@@ -44,13 +44,19 @@ def detect_api_version(env)
...
-    def set_api_version(env, version)
+    def set_api_version(env, version, variant)
       # API modules will look for the version in env
-      env["HTTP_X_API_VERSION"] = version
+      env["api.version"] = version
+      env["api.variant"] = variant
+    end
+
+    def set_log_context(version, variant)
+      Pliny.default_context[:api_version] = [version, variant].join(".")
```
gudmundur: This should probably be merged into `Pliny::RequestStore.store[:log_context].merge!`. `Pliny.default_context` is for application specific things like app name and deployment.

lib/pliny/commands/generator/schema.rb
dmathieu: No. `"#{schema_path}/schema.json"` would become `./docs/schema/schema.json`

lib/pliny/commands/generator/schema.rb
dmathieu: No, because one needs to be `./docs/schema.json`, and the other one `./schema/schema.json`.
We need the conditional to handle the file being inside the schema folder or not.

lib/pliny/middleware/versioning.rb
dmathieu: `Pliny::RequestStore.store[:log_context].merge!(api_version: [version, variant].join("."))`?

lib/pliny/commands/generator/schema.rb
gudmundur: Ah! Thanks.

lib/pliny/middleware/versioning.rb
dmathieu: That's updated.

lib/pliny/middleware/versioning.rb
```diff
@@ -44,13 +44,21 @@ def detect_api_version(env)
...
+    def set_api_version(env, version, variant)
       # API modules will look for the version in env
-      env["HTTP_X_API_VERSION"] = version
+      env["api.version"] = version
+      env["api.variant"] = variant
+    end
+
+    def set_log_context(version, variant)
+      Pliny::RequestStore.store[:log_context].merge!(
+        api_version: [version, variant].join(".")
```
gudmundur: If no version or variant is specified, this will add `api_version=3.` to the log context. Note the optional `.`.

gudmundur:
I've been testing this locally as this is a pretty big change. Generating schemas works like a charm, and so do the router helpers. Though I've been unable to get it working with `curl` invoking a version variant endpoint. Would you mind jumping on a quick screen share to figure out what's going on?

lib/pliny/middleware/versioning.rb
dmathieu: Thank you. I've fixed this.

gudmundur:
@dmathieu Sorry for leaving this one hanging for so long. Even though this looks good, I'm a little reluctant bringing it in, as we've been discussing OOB whether we should switch over to OpenAPI, but I've also seen on Slack people discussing GraphQL. Before bringing this in, I'd love to get hear what @brandur and @pedro think about variants in the context of swapping JSON schemas out, or whether we should make it pluggable.

pedro:
I think this looks great in the context of what Pliny does today and JSON schema and all, so +1 for merging.

In parallel I'd also be a proponent of nuking JSON schema. I love the benefits we get from it – like param/response validation, stubs, docs, etc – but no one needs to write raw JSON/Yaml to get these today. Life is better when you specify the API in the same language you implement it; can't tell how many times I've went down a rabbit hole trying to figure out how to reuse/reference stuff in the schema, which is never a problem when you use an actual programming language to define it instead.

I initially wrote [sinatra-schema](https://github.com/pedro/sinatra-schema) as an experiment defining schemas in Ruby, but these days am pretty happy just using GraphQL. Up to you which direction to take with Pliny of course – I can only speak for my limited but generally positive experience with it 😬

geemus:
@pedro yeah, I think some of the challenge comes from if/when you need cross-language support (which in theory is easier with a neutral thing like json). In practice though, as you suggest, it kind of ends up meaning that everybody just needs to learn schema stuff, which might as well be a language anyway.

pedro:
@geemus wessss! ❤️  right on, JSON schema is still great for defining some API once and implementing in any language/framework. We'd lose that going with GraphQL/others, but still a worthwhile move I suspect.

geemus:
@pedro yeah, in theory it all sounds good for interoperability, but in practice it is WAY too much friction.

brandur:
I think that variants had some promise, but practically speaking might be better implemented as something like user-specific gating (that's what we do at Stripe). It also seems to me to be a very good idea to keep the number of total versions of your API down, and that includes things like variants.

Regarding JSON hyper-schema: in the short term, nuking it for OpenAPI seems like the right solution to me for two reasons:

1. The two formats are fairly similar, and OpenAPI makes significant use of JSON schema to describe resources and responses. I think it would be relatively straightforward to write a script to migrate people's existing hyper-schemas to OpenAPI.
2. JSON hyper-schema seems like a dead evolutionary branch of APIs. As far as I can tell, Heroku and a few others using Pliny are the only organizations really using it.

Longer term, I get the feeling that there's a growing sentiment that RESTful APIs in general are snake oil and not worth all the fuss that's being made about them. In my opinion, the paradigm's promises around graceful progression through techniques like HATEOAS and redirecting clients with 301s was an aspirational promise that never materialized. As a provider, you can't know whether your clients are integrating correctly, and unless you're providing a very low value service, you just can't risk breaking them by messing around with this type of change.

Don't throw out your RESTful API if you have one, but the future is probably something closer to GraphQL, and it would be neat to consider how that fits into Pliny.

geemus:
@brandur agreed, variants were better than nothing but I think end up being pretty problematic in their own right (they might as well just be additional/distinct versions).

1. I suspect that migration as you mention would be pretty tenable and have considered but not yet pursued (I'd like to get out of the business of maintaining unique/independent tooling and just use OpenAPI stuff if possible).
2. Hyper-schema has seen a [fork](https://github.com/json-schema-org/json-schema-spec) and some resurgence of effort on the spec itself, but OpenAPI has much broader buy-in and greater momentum.

Agreed that I also have a hard time seeing my way to a place where I can confidently make changes and not expect it to break things. I also wonder where things might be headed in that regard, whether GraphQL (which has it's own issues) or elsewhere.

brandur:
> Hyper-schema has seen a fork and some resurgence of effort on the spec itself, but OpenAPI has much broader buy-in and greater momentum.

Oh interesting. I didn't know about the fork!

> I also wonder where things might be headed in that regard, whether GraphQL (which has it's own issues) or elsewhere.

GraphQL has the same backward incompatibility problems, but things are made a little better because at least clients have to specify which fields they use. You could imagine a world where you could empirically verify that no one was left using some deprecated field and remove it completely. My analysis is that practically speaking it's no worse than what you'd see in a REST API, and probably slightly better.

PR #261, opened 2016-04-13
Upgrade sinatra and sinatra-contrib

gudmundur:
Older versions than 1.4.7 of sinatra-contrib raise a warning on Ruby 2.3 due to a deprecation of Thread.exclusive. See https://github.com/sinatra/sinatra-contrib/pull/184. This issue has been leaving me reluctant to cut a new release of Pliny, as every request in development would output `Thread.exclusive is deprecated, use Mutex` and a stack trace. Eew.


brandur:
Nice find :) +1.

PR #262, opened 2016-04-18
Upgrade rack-timeout

gudmundur:
Upgrades to `Rack::Timeout` v0.4 by request of @kch. The new version uses an initializer for setting the default timeout as well as using monotonic time.


dmathieu:
:+1:

brandur:
Love it! +1.

schneems:
I'm getting a failure

```
ArgumentError: comparison of String with 0 failed
```

Is there some way to specify that this should be an integer in the config? 

Here's my config:


```
  override \
    db_pool:          5,
    timeout:          45,
``

Issue #92, opened 2014-11-05
Pliny exceptions getting logged to Rollbar


pedro:
Rollbar is injecting its Rack::Builder middleware in a way that it's getting precedence over `Pliny::Middleware::RescueErrors`. This means that internal Pliny exceptions like `Unauthorized` are getting logged before we rescue and convert them into a proper http response.

kennyp:
@pedro Looks like they merged your PR. What do we need to change to get this issue closed?

pedro:
Nothing :} @brandur nailed this down on 1dba7342329ab084e8d742dce6d3decebf3911dc, happy to report I couldn't reproduce the original issue on a new Pliny app.

Thanks!

PR #264, opened 2016-04-21
Disable Rollbar monkey patching

gudmundur:
During investigation of an incident I noticed that Pliny has been double posting to Rollbar for a while. I traced this to a missing `disable_monkey_patch` as a part of #110 a while back.

**Note** Bringing this change in might have the side effect of error reporting being disabled in certain places, for example in Sidekiq workers and other `bin/` scripts.

### DIFF
--- a/lib/template/config/initializers/rollbar.rb
+++ b/lib/template/config/initializers/rollbar.rb
@@ -1,5 +1,6 @@
 Rollbar.configure do |config|
   config.enabled = ENV.has_key?('ROLLBAR_ACCESS_TOKEN')
+  config.disable_monkey_patch = true
   config.access_token = ENV["ROLLBAR_ACCESS_TOKEN"]
   config.environment = ENV['ROLLBAR_ENV']
   config.logger = Pliny::RollbarLogger.new


gudmundur:
I'll be opening a PR into `pliny-sidekiq` shortly to enable Rollbar reporting by default if it hasn't been activated.

brandur:
Oh no! Too bad this seems to have been lost, but thanks a lot for tracking it down! +1.

> I'll be opening a PR into pliny-sidekiq shortly to enable Rollbar reporting by default if it hasn't been activated.

Good call.

pedro:
wow, sorry I missed this :+1: :+1: :+1:

gudmundur:
@brandur @pedro To refresh my memory before bringing this in: What is the motivation for disabling monkey patching? If Pliny is biased towards Rollbar, wouldn't we want it enabled by default everywhere? Even though it means monkey patching? If middleware ordering is all we care about, there's now the option of just disabling monkey patching for rack, `disable_rack_monkey_patch`.

pedro:
@gudmundur middleware ordering is all we care about, [original issue here](https://github.com/interagent/pliny/issues/92). I didn't know of `disable_rack_monkey_patch`, this seems like a better fit then!

brandur:
+1. I think Pedro submitted the original monkey patch option, so the Rack
version must be new since then.

gudmundur:
Oooook, been diving into this some more and what I'll do is this:

1. Use `disable_rack_monkey_patch`.
2. Remove the Rollbar middleware from `lib/routes.rb`.
3. Add explicit Rollbar handling to the `RescueErrors` middleware. Basically inlining the middleware.

My reasoning here, is that `RescueErrors` will gobble up all exceptions by translating them to `Pliny::Errors::InternalServerError`, unless `RAISE_ERRORS` is enabled. The downside is that anything further up the middleware chain will not be reported to Rollbar.

gudmundur:
Closing in favor of #265.

PR #265, opened 2016-05-02
Rollbar and RescueErrors revamp

gudmundur:
In a comment on #264 I described a solution for handling exceptions raised within a request. When I looked into the Heroku API, it turns out that's exactly the approach I was considering. Much of this code is lifted from there, and is the work of @brandur.

### DIFF
--- a/lib/pliny/error_reporter.rb
+++ b/lib/pliny/error_reporter.rb
@@ -1,0 +1,42 @@
+require 'rollbar/exception_reporter'
+require 'rollbar/request_data_extractor'
+
+class Pliny::ErrorReporter
+  def self.notify(exception, context: {}, rack_env: {})
+    Pliny.log_exception(exception)
+
+    REPORTERS.each do |reporter|
+      begin
+        reporter.new.notify(exception, context: context, rack_env: rack_env)
+      rescue
+        Pliny.log_exception($!)
+      end
+    end
+  end
+
+  class RollbarReporter
+    include ::Rollbar::ExceptionReporter
+    include ::Rollbar::RequestDataExtractor
+
+    def notify(exception, context:, rack_env:)
+      Rollbar.reset_notifier!
+      scope = fetch_scope(context: context, rack_env: rack_env)
+      Rollbar.scoped(scope) do
+        report_exception_to_rollbar(rack_env, exception)
+      end
+    end
+
+    private
+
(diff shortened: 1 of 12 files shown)


lib/pliny/middleware/rescue_errors.rb
```diff
@@ -8,25 +8,12 @@ def initialize(app, options = {})
     def call(env)
       @app.call(env)
     rescue Pliny::Errors::Error => e
-      # blank out this field so that the error is not picked up by Rollbar
-      env["sinatra.error"] = nil
-
       Pliny::Errors::Error.render(e)
     rescue Exception => e
```
joshwlewis: Rescuing `Exception` and not re-raising it causes interrupts and sigkills to be ignored. Rescuing `StandardError` might make more sense here, or alternatively, re-raising if the error is not a `StandardError`.

lib/pliny/middleware/rescue_errors.rb
```diff
@@ -8,25 +8,12 @@ def initialize(app, options = {})
     def call(env)
       @app.call(env)
     rescue Pliny::Errors::Error => e
-      # blank out this field so that the error is not picked up by Rollbar
-      env["sinatra.error"] = nil
```
joshwlewis: Could we log this error, or at least that we are rescuing it? It might be handy to track down user-reported errors.

joshwlewis: Looking again, the instrumentation middleware probably does this well enough.

gudmundur:
@brandur nudge nudge

lib/pliny/error_reporter.rb
```diff
@@ -0,0 +1,42 @@
...
+      {
+        request: proc { extract_request_data_from_rack(rack_env) }
+      }
+    rescue Exception => e
+      report_exception_to_rollbar(rack_env, e)
+      raise
+    end
+  end
+
+  REPORTERS = [RollbarReporter]
```
brandur: It might make sense to make this configurable at some point in the future, but not too big of a deal as long as there's still one reporting option.

It's possible too that we'll establish a convention for this with the metrics reporter.

brandur:
@gudmundur Hey, sorry about the delay! This looks good to me. Nice work :) Ship it!

PR #267, opened 2016-05-16
Config cleanup

gudmundur:
* Removes `Config.placeholder`, not referred to anywhere in Pliny. Not sure what the intention even is.
* Changes the default `Config.pliny_env` to `production`. `.env.sample` and `.env.test` do the right thing already.


dmathieu:
:+1:

brandur:
+1!

PR #268, opened 2016-05-16
Add #count and #measure methods to Pliny

gudmundur:
Adds `Pliny::Metrics.count(*names, value: 1)` and `Pliny::Metrics.measure(*names, &block)` methods. By invoking the methods, they will emit logfmt entries.

### Examples 

* `Pliny::Metrics.count(:hello)` will emit `count#my_pliny_app.hello=1`.
* `Pliny::Metrics.measure(:hello) { sleep(1) }` will emit `measure#my_pliny_app.hello=1.000`.


lib/pliny/metrics.rb
```diff
@@ -0,0 +1,20 @@
...
+      counts = Hash[names.map { |n| ["count##{Config.app_name}.#{n}", value] }]
+      Pliny.log(counts)
+    end
+
+    def measure(*names, &block)
+      start = Time.now
+      return_value = block.call
+      elapsed = Time.now - start
+
+      measures = Hash[names.map { |n| ["measure##{Config.app_name}.#{n}", elapsed] }]
```
dmathieu: This is slightly obscure, and the exact same thing as in `#count`. How about extracting it into another method? Something like:

```
def monitor(key, names, value)
  data = Hash[names.map { |n| ["#{key}##{Config.app_name}.#{n}", value] }]
  Pliny.log(data)
end
```

gudmundur: If I add it as a private on the module, won't it still contaminate the `Pliny` namespace? That was my main motivator for not extracting a method here.

brandur: If you want to extract it, you may be able to try some other tricks. e.g. Creating a nested `Helpers` module within `Metrics` that gets referenced by these implementations, but which doesn't get mixed into `Pliny`.

dmathieu: That might be over-extracting, but there could be another module for that method so we don't pollute the main class.

_fairly bad idea (or not): we could also prepend the method name with two underscores_

brandur:
Good stuff! Two thoughts:

1. It may be worth thinking about making the metrics module configurable so that there could be more future options beyond just emitting log lines. I think most of the Heroku/Pliny practices are reasonable enough to emulate elsewhere, but I've always thought of the log-based emission somewhat contentious just because it's quite inefficient/expensive compared to say a UDP-based solution for example.
2. Should we force the metrics helpers to be referenced at `Pliny::Metrics` rather than mixing them into the top level namespace? It's a little more typing, but I don't think the explicitness is a bad thing.

gudmundur:
@brandur I've moved the methods into `Pliny::Metrics`. As for abstracting away how the metrics are emitted, I'd prefer to punt on that until that use case actually comes up. Is that alright with you?

brandur:
> @brandur I've moved the methods into Pliny::Metrics. As for abstracting away how the metrics are emitted, I'd prefer to punt on that until that use case actually comes up. Is that alright with you?

@gudmundur Yep, absolutely! It seems like the interface should be generic enough that it won't back us into a corner.

LGTM! Sorry about the review delay. Feel free to bump these more often if I'm not answering because it might mean that I declared inbox zero on GitHub issues :)

PR #269, opened 2016-05-20
upgrade to Ruby 2.3.1 + update pinned gems

bgentry:
Followed what was done in #253. Puma has a 3.4.0 out now but thought I should only move this to the point release.


brandur:
Thanks Blake!

IMO we should be pretty aggressive on the gem updates (including going to Puma 3.x). The Gemfile here is only for the template, so all existing Pliny apps should stay pretty stable in that they can upgrade on their own schedules.

gudmundur:
@bgentry I've just cut a release for this, `v0.17.0`.

PR #272, opened 2016-05-27
Update apps by reapplying the templates files

### LINKED ISSUE #271: pliny-update copies raw erb files
Ran into this problem upgrading telex. The .env and app.json files were copied over from the template instead of being evaluated. 
I looked into submitting a patch, but they need the app name and I'm not entirely sure the best way to retrieve that on an upgrade.

dmathieu:
We need to apply ERB when updating, which cannot reliably be done with a diff content.
Anyone updating can still check it's changes by inspecting the unstaged changed in their repository.

cc @dmcinnes @geemus 
Closes #271

### DIFF
--- a/lib/pliny/commands/updater.rb
+++ b/lib/pliny/commands/updater.rb
@@ -5,7 +5,7 @@ require 'pliny/commands/creator'
 require 'uri'
 
 module Pliny::Commands
-  class Updater < Creator
+  class Updater
     attr_accessor :stream
 
     def self.run(stream = $stdout)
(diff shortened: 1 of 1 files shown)


geemus:
Probably don't want the `.DS_Store` files there.

So if I understand/read correctly, this just leaves the erb output-ed files as-is and simply updates templates? Seems reasonable.

dmathieu:
Yes. The changes were previously available in a diff file. They will now be available with `git diff`.

geemus:
Sounds good

dmathieu:
I've rebased this branch to let the tests pass now that #273 is merged.

lib/pliny/commands/updater.rb
```diff
@@ -1,10 +1,11 @@
 require 'fileutils'
 require 'pathname'
 require 'pliny/version'
+require 'pliny/commands/creator'
 require 'uri'
 
 module Pliny::Commands
-  class Updater
+  class Updater < Creator
```
gudmundur: Inheriting from the `Creator` looks a little strange from me. Was there any specific decision to do this, other than avoid duplicating a few methods? My worry is that there might be unintended side effects by this if something changes in the `Creator`.

gudmundur:
Thanks @dmathieu. This seems to fix an issue that I just got nudged about. I've added a comment on the change, but otherwise looks good to me.

lib/pliny/commands/updater.rb
dmathieu: I think I did this to avoid some code repetition. I'm fine removing it though, I just did so.

lib/pliny/commands/updater.rb
```diff
@@ -1,6 +1,7 @@
 require 'fileutils'
 require 'pathname'
 require 'pliny/version'
+require 'pliny/commands/creator'
```
gudmundur: I'm assuming this line is not needed anymore?

PR #273, opened 2016-06-01
Use rspec-mocks instead of rr

dmathieu:
RR just broke when releasing 1.2.0. This switched to rspec-mocks.
The gem is in our dependency tree anyway, as it's added by rspec.

PR #274, opened 2016-06-01
Make Rollbar dependency optional

fdr:
Previously, one could use Pliny with no error tracker or other error
trackers, but in 0.17.0 requiring Pliny will crash if one has removed
Rollbar.

To fix this, use a conditional class definition to fix this.

Incidentally, I'm still testing this out, but running the specs is broken from missing mocking symbols, per #273...so I'm abusing the Travis buildfarm to test.

### DIFF
--- a/lib/pliny/error_reporter.rb
+++ b/lib/pliny/error_reporter.rb
@@ -1,7 +1,6 @@
-require 'rollbar/exception_reporter'
-require 'rollbar/request_data_extractor'
-
 class Pliny::ErrorReporter
+  REPORTERS = []
+
   def self.notify(exception, context: {}, rack_env: {})
     Pliny.log_exception(exception)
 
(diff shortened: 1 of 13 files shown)


lib/pliny/error_reporter.rb
```diff
@@ -14,29 +13,35 @@ def self.notify(exception, context: {}, rack_env: {})
     end
   end
 
-  class RollbarReporter
-    include ::Rollbar::ExceptionReporter
-    include ::Rollbar::RequestDataExtractor
+  begin
```
brandur: I'm personally not super crazy about rescuing `LoadError` on require attempts. I wonder if we can make this conditional based on the current error reporting configuration?

fdr: Maybe? I didn't try even one other approach, I mostly wanted to make sure that this was the only problem with rollbar being optional. So far, it seems to be the case.

Maybe it could live in a templated initializer?

lib/pliny/error_reporter.rb
```diff
@@ -14,29 +13,35 @@ def self.notify(exception, context: {}, rack_env: {})
...
+    require 'rollbar/exception_reporter'
+    require 'rollbar/request_data_extractor'
 
-    def notify(exception, context:, rack_env:)
-      Rollbar.reset_notifier!
-      scope = fetch_scope(context: context, rack_env: rack_env)
-      Rollbar.scoped(scope) do
-        report_exception_to_rollbar(rack_env, exception)
+    class RollbarReporter
```
dmathieu: I'm not a big fan of having this in the main error_reported file.
How about having `lib/pliny/error_reporters/rollbar`?

dmathieu:
Yay for rspec-mock.

lib/pliny/error_reporter.rb
fdr: okay, but what if not blowing up if rollbar is not installed? How does it get toggled off?

In the case of putting this code in a initializer put into the user's project, it's pretty simple: delete the initializer. 

The downside is that the more stuff in the template, the harder it is to track upstream additions to functionality.  I already have to do small backports to update Pliny versions.

 Perhaps it can live in a file that is "required" in the template explicitly to get the best of both worlds, so non-rollbar users can delete that "require" line?

lib/pliny/error_reporter.rb
brandur: +1. The best I can up with is moving the reporter to its own file as suggested below, and then just load and configure it from an initializer:

``` ruby
require 'lib/pliny/error_reporters/rollbar'
Pliny::ErrorReporters << Pliny::RollbarLogger.new

Rollbar.configure do |config|
  config.enabled = ENV.has_key?('ROLLBAR_ACCESS_TOKEN')
...
```

(Or something.)

gudmundur:
@fdr Excellent catch. I'm inclined to agree with @brandur on the `LoadError` and his suggestion for doing the require in the `config/initializers/rollbar.rb`. That way if you nuke initializer, it'll never bring it in. What do you think?

fdr:
On Fri, Jun 10, 2016 at 7:23 AM Guðmundur Bjarni Ólafsson <
notifications@github.com> wrote:

> @fdr <https://github.com/fdr> Excellent catch. I'm inclined to agree with
> @brandur <https://github.com/brandur> on the LoadError and his suggestion
> for doing the require in the config/initializers/rollbar.rb. That way if
> you nuke initializer, it'll never bring it in. What do you think
>

Sounds perfect to me, I just haven't had the time to do it myself and find
a/the spot for optional pliny require-able modules.

gudmundur:
@fdr As it was me that initially broke this, I whipped up another PR to resolve this, check it out #275.

gudmundur:
@fdr Closing this as #275 has just been merged.

PR #275, opened 2016-07-09
Make Rollbar optional

gudmundur:
After bringing in #265, Rollbar became by accident a mandatory dependency of Pliny. This was pointed out by @fdr in #274. The approach in #274 was to rescue on `LoadError` if Rollbar was not found. With the approach in this PR, if the `config/initializers/rollbar.rb` file is removed, Rollbar will not be loaded at all.

cc: @fdr @brandur


fdr:
I played with this and it seems to work.

brandur:
Sorry about the really delayed response, but this looks amazing @gudmundur! +1.

PR #276, opened 2016-08-23
Allow handler to specify custom message for uncaught errors

uhoh-itsmaciek:
Thoughts on this? I know I should handle all potential errors explicitly, but if something sneaks through, it'd be nice to provide a more friendly message than "Internal server error."

### DIFF
--- a/lib/pliny/middleware/rescue_errors.rb
+++ b/lib/pliny/middleware/rescue_errors.rb
@@ -3,6 +3,7 @@ module Pliny::Middleware
     def initialize(app, options = {})
       @app = app
       @raise = options[:raise]
+      @message = options[:message]
     end
 
     def call(env)
@@ -13,7 +14,7 @@ module Pliny::Middleware
       raise if @raise
 
       Pliny::ErrorReporters.notify(e, rack_env: env)
-      Pliny::Errors::Error.render(Pliny::Errors::InternalServerError.new)
+      Pliny::Errors::Error.render(Pliny::Errors::InternalServerError.new(@message))
     end
   end
 end
--- a/spec/middleware/rescue_errors_spec.rb
+++ b/spec/middleware/rescue_errors_spec.rb
@@ -43,12 +43,22 @@ describe Pliny::Middleware::RescueErrors do
     end
   end
 
+  it "uses a custom error message with the message option" do
+    @app = new_rack_app(message: "Please stand by")
+    allow(Pliny::ErrorReporters).to receive(:notify)
+    get "/"
+    assert_equal 500, last_response.status
+    error_json = MultiJson.decode(last_response.body)
+    assert_equal "internal_server_error", error_json["id"]
+    assert_equal "Please stand by", error_json["message"]
+  end
+
   private
 
   def new_rack_app(options = {})
     Rack::Builder.new do
       use Rack::Lint
-      use Pliny::Middleware::RescueErrors, raise: options[:raise]
+      use Pliny::Middleware::RescueErrors, raise: options[:raise], message: options[:message]
       run BadMiddleware.new
     end
   end


brandur:
@uhoh-itsmaciek Sorry about the slow response Maciek! My personal inbox unfortunately had become a little overflown.

This seems like a reasonable/useful change to me. I'd ask anyone else who's seen this to point out any other callbacks that they remember which we might have in the project that are already named in a way other than the `on_` prefix proposed here (in case we want to follow an alternative convention instead). That said, I think Puma uses `on_` in its configuration, so it works for me.

+1.

uhoh-itsmaciek:
NBD, as Spiderman would say. For now, I copied the existing error handling middleware into our project. For what it's worth, I don't think `on_error` is a great name, since it suggests an error *handler*: this is really more about more flexibility in error *messages*, and doesn't really have the power to alter the error handling logic (nor should it). But I couldn't think of a better name, so here it is. I welcome more bikeshedding / brainstorming on the name.

brandur:
@uhoh-itsmaciek Yeah, good point. On further thought, just providing flexibility in the error _message_ rather than the entire error itself strikes me as a little binding. What do you think about the callback specifying the entire error raising behavior? Users should be able to pretty easily throw their own `Pliny::Errors::InternalServerError` is so desired.

uhoh-itsmaciek:
I mean, that works for me (and in fact what I implemented originally), but I submitted the PR like this because I couldn't come up with a use case for the more general behavior: it seems like unknown errors shouldn't get to the top-level handler in the first place. Still, it's not totally crazy to allow more flexibility there, and happy to rework the PR to do that if you prefer (though can `Pliny::Errors::Error.render` render an arbitrary error? If not, what do we do if the handler returns a non-Pliny error? If it returns something that's not an error at all?

brandur:
@uhoh-itsmaciek Hm, good points. Despite what I said above, if we do go down the more flexible path, we might want to include the call to `Pliny::Errors::Error.render` _inside_ the callback as well just so that it could all be used purely generically.

Another one we could try is just rename the callback to `on_error_message` or something of the like so that in case we didn't introduce a more flexible path later, we'd at least have a sane naming scheme.

I'm pretty easy though, so I'd certainly defer to the group's will on this one.

cc @pedro @fdr Any thoughts on this one? As users, what would you prefer?

uhoh-itsmaciek:
Sure. @gudmundur had also suggested just doing a static string (in which case it could be a class method). I think it might be too restrictive in general, but it would work for my use case, so I'm fine with it. Would love to hear others' thoughts and happy to update the PR to any of these options.

fdr:
If I'm going to bother making error messages at all, I'd prefer control over what's in it to provide some detail.

pedro:
I think `en.yml` is the best way to define copy.

We're doing that in the Lugg API in fact – error messages are in that file:

```
en:
  errors:
    internal_server_error: Foobar
```

And we reuse that setup to define error messages for our custom errors too (eg: when we render a 422 with a special id/message pair to the client).

I can port that in if this seems like a good fit.

gudmundur:
@pedro Oh nice. I like this approach. It is restricted in that you can't override it with code, which is acceptable to me. When you're in the situation of a `500 Internal Server Error`, it's fairly limited what the user can do about that, other than try again.

pedro:
Right on, it's not as flexible but you can always define have a custom error class if you want to further customize the error handling.

fdr:
How does one normally internationalize a ruby and/or sinatra application? It seems anemic to me that one would localize error code messages specially in this way but not also want to integrate a translation toolchain, e.g. gettext, or whatever.

pedro:
Oh yes we'd integrate the standard Ruby localization stuff with `I18n`:

http://guides.rubyonrails.org/i18n.html

So it can be reused for any messages rendered to users really.

gudmundur:
@uhoh-itsmaciek Did you end up doing something in your own apps? I'd love to see something relating to this land.

uhoh-itsmaciek:
@gudmundur I'm running a vendored fork of the middleware and passing a fixed error message [here](https://github.com/interagent/pliny/pull/276/files#diff-a913f204f5f1dab9d9c618d7610c5095R17). I'm happy to change the PR to accept a fixed error message if that sounds reasonable.

gudmundur:
@uhoh-itsmaciek yeah, I think if your use case is a fixed error message, then lets stick with that for now. We can always add support for a proc later on.

uhoh-itsmaciek:
@gudmundur done, force-pushed a new version to the same branch.

PR #279, opened 2016-10-04
Clear request store before logging request

gudmundur:
While debugging an issue the other day, I noticed that the `Pliny::Middleware::Instruments` was logging out the `log_context` from a previous request. Diving further into it, it occurs because of ordering of loading middleware. There are two solutions that quickly come to mind to address this issue:

1. Re-ordering the middlewares so `Pliny::Middleware::RequestStore` runs before `::Instruments`. This is what this PR does and is the simplest solution. There are currently two middlewares before `::RequestStore`, mainly as this middleware has a dependency on the `::RequestId` middleware. The downside here is that if any of the earlier middlewares start using logging, we're in the same spot as before.

2. Split the `::RequestStore` middleware into two. `::RequestStore::Clear`, which would be the very first middleware and `::RequestStore::Seed` which would be loaded after any dependencies. This is what the Heroku API currently does. This means introducing breaking change, but given the importance of fixing this, I'm ok with.

I'd appreciate advisement on this.

### DIFF
--- a/lib/template/lib/routes.rb
+++ b/lib/template/lib/routes.rb
@@ -1,9 +1,9 @@
 Routes = Rack::Builder.new do
   use Pliny::Middleware::CORS
   use Pliny::Middleware::RequestID
+  use Pliny::Middleware::RequestStore, store: Pliny::RequestStore
   use Pliny::Middleware::Instruments
   use Pliny::Middleware::RescueErrors, raise: Config.raise_errors?
-  use Pliny::Middleware::RequestStore, store: Pliny::RequestStore
   use Rack::Timeout,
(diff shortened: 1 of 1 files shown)


brandur:
Oops, I should've seen this coming (as I think I may have written that patch for the Heroku API).

I guess the main thing that I don't like about the patch as currently written is that it's far too easy to achieve a regression in that someone may reorder the middleware in the future and there is no mechanism (testing or otherwise) that will catch it.

I think that splitting into `::Clear` and `::Seed` makes this at least a little bit better in that you can add a big comment over `::Clear` to lesson the danger slightly. There may be other ways to go about this as well though. Thoughts?

PR #280, opened 2016-10-05
Split request store middleware

gudmundur:
This is an alternate solution to a another PR that I opened, see [here for more context](https://github.com/interagent/pliny/pull/279).

This solution splits the `RequestStore` middleware into a `RequestStore::Clear`, which will be the first middleware, and then `RequestStore::Seed` immediately after the `::RequestId` middleware. By bumping it up two places, the `::Instruments` middleware doesn't need to worry about `request_ids` as they come from `log_context`.

This also contains a couple of fixes where test doubles are leaking across tests through doubles.


gudmundur:
@brandur here's the alternate solution, which the Heroku API uses.

brandur:
@gudmundur Crazy fast work here. Nice!

Looks good to me. +1 for bringing mainline.

There are of course some backwards compatibility issues here and it would be nice to do what we can to make sure that people aren't too surprised by the change. We should probably start maintaining a change log or something ...

gudmundur:
@brandur A changelog is precisely what I'm starting as well. I've been kind of maintaining one in the Releases here on GitHub, I'll migrate that over to a `CHANGELOG.md`.

brandur:
> A changelog is precisely what I'm starting as well. I've been kind of maintaining one in the Releases here on GitHub, I'll migrate that over to a CHANGELOG.md.

Good call! Lemme know if you want any help with that.

Issue #277, opened 2016-08-26
Make use of APP_ENV


gudmundur:
Sinatra recently merged a change which adds a new `APP_ENV`. It probably makes sense to deprecate `PLINY_ENV` and use this instead.

See: https://github.com/sinatra/sinatra/pull/984

brandur:
@gudmundur +1. Good call.

Strangely, I also happened to stumble across that exact Sinatra issue a few days ago.

PR #282, opened 2016-10-19
Replace PLINY_ENV with APP_ENV

soulim:
Resolves https://github.com/interagent/pliny/issues/277.

`PLINY_ENV` and `pliny_env` have been replaced with `APP_ENV` and `app_env` everywhere in the code.

In order to help with transition, a deprecation message is displayed if `Config.pliny_env` is used. So `pliny_env` is actually deprecated and can be still called with a grumpy message. Deprecation message is based on [the pattern used in Sinatra](https://github.com/sinatra/sinatra/blob/1b0edc0aeaaf4839cadfcec1b21da86e6af1d4c0/lib/sinatra/base.rb#L945).

Would be happy to hear your thoughts on this one 😄

### DIFF
--- a/lib/pliny/config_helpers.rb
+++ b/lib/pliny/config_helpers.rb
@@ -50,19 +50,21 @@ module Pliny
     end
 
     def rack_env
-      if app_env == "development" || app_env == "test"
+      if env == "development" || env == "test"
         "development"
       else
         "deployment"
       end
     end
-    
-    # DEPRECATED: PLINY_ENV is deprecated in favour of APP_ENV.
+
+    # DEPRECATED: pliny_env is deprecated in favour of app_env.
     #             See more at https://github.com/interagent/pliny/issues/277
+    #
+    # This method is kept temporary in case it is used somewhere in the app.
     def pliny_env
       warn "Config.pliny_env is deprecated and will be removed, " \
            "use Config.app_env instead."
-      app_env
+      env
     end
 
     private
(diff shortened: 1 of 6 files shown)


lib/pliny/config_helpers.rb
```diff
@@ -50,12 +50,20 @@ def array(method = nil)
     end
 
     def rack_env
-      if pliny_env == "development" || pliny_env == "test"
+      if app_env == "development" || app_env == "test"
```
brandur: I wonder if we should respect the value of `app_env` but _also_ `pliny_env` here for a bit longer to give people some opportunity to upgrade?

I'm a little worried that with this if someone bumped their Pliny version behavior based off `rack_env` would change without warning in their environments because they hadn't injected the new `APP_ENV` variable into `.env` and/or their app's config.

Thoughts?

soulim: Yeah... Now when you said that, I also think backward compatibility with `pliny_env` could be better. I'll improve the solution and will add more tests on cases when `PLINY_ENV` is still defined in the environment.

Thanks for the idea! 👍

brandur:
Excellent! Thank you :)

soulim:
Updated solution according to our discussion 😄  In case `PLINY_ENV` is still around, it will be used, but with appropriate warnings.

Additionally I've added short comments to new methods. I hope they would be helpful for other developers.

What do you think?

brandur:
> Additionally I've added short comments to new methods. I hope they would be helpful for other developers.

Love it. Thanks!

> What do you think?

Looks good to me. Pulling this in.

PR #283, opened 2016-10-25
Add authorization helpers

soulim:
Similar to `Pliny::Helpers::Serialize` authorization helpers define public interface that might be used to set and call an authorizer on specific or all endpoints.

Examples

``` ruby
module Authorizers
  class Token
    def initialize
      @token = 'foo'
    end

    def authorized?(request)
      @token == request.params['access_token']
    end
  end
end

module Endpoints
  class Foo < Base
    authorizer Authorizers::Token

    before { authorize! }

    get do
    end
  end
end
```

Authorizers are PORO that must respond to `#authorized?` predicate method. The method receives a request object as a parameter and returns true or false depending on results of the authorization. Because of that all authorizers can be easily tested 😄 

Additional bonus: it is possible to define a global authorizer and special authorizers for specific endpoints. Actually this is exactly how the idea of authorization helpers was born 😄 

I understand that different serializers for each endpoint is much common than authorizers. Anyway I would like to share the idea and would be happy to hear any feedback.


brandur:
Hi @soulim,

Sorry as usual for the delay here. Thanks a lot for the PR and the well-written description! I really like the idea of keeping these as POROs.

So first of all, I might suggest that "authenticator" might be a better name for this feature given that we're mostly trying to perform identity verification ([see this SO post on authentication versus authorization](http://stackoverflow.com/questions/6556522/authentication-versus-authorization)).

I see that you've included a basic test for authenticating against a token, but I'm just trying to think of what's probably the more standard case where you'd have an incoming token and want to figure out what user, if any, that token belongs to. How would the authorizer get that user back to the API endpoint?

It would also be more ideal IMO if we could abstract away the `authorize!` endpoint. It seems like in this case the `authorizer` directive should be enough indication on the part of an endpoint author that the given class could be run.

It might also be worth thinking about how this could potentially interact with a resource access control system that could conceivably be introduced at some point in the future. Should that kind of concerned be mixed into these classes, or should that be a totally separate part of the framework?

Would definitely love to hear some feedback from any other users who might have opinions on this too! Would this be a feature that you'd find useful?

gudmundur:
I'm working in related areas of the Pliny projects I've got. What I'm specifically doing is adding another method of authenticating, so in one or two projects we'll support three different auth backends. I'm also figuring out how to tie that into authorization, which happens by making API calls to another service. What I'm trying to get at is a where I can extract the authorization bit out into a Gem, while leaving authentication in the caller. Once I've gotten to that point, I'm sure I'll have some better input on this PR on how this might be generalized.

PR #285, opened 2016-11-25
Add and apply rubocop to template

cyberdelia:
This add and apply [rubocop](http://rubocop.readthedocs.io) to the Pliny template. It also use ruby 2.3.3 for tests and the template.

I disabled cops that where not trivial to correct, or did not provide auto-correction. Feel free to modify .rubocop.yml or the code beyond what I did.


pedro:
Very nice, +999!

The only thing I'd change are strings, is there a new recommendation somewhere to go with single quoted ruby strings? I thought it was double-quoted all the way down 😬

brandur:
Yeah, this definitely seems like a good idea. Thanks for the work here @cyberdelia!

> The only thing I'd change are strings, is there a new recommendation somewhere to go with single quoted ruby strings? I thought it was double-quoted all the way down 😬

+1. I've also never seen much of a concrete reason for using the single quoted stuff, but it has at least a few downsides. Mostly that it takes some time to move over to double quotes if you realize that you need interpolation, but I've also seen it prompt people to write specs like this:

``` ruby
it 'doesn\'t throw an exception' do
  ...
```

Which is a little silly to say the least.

gudmundur:
Personally I use single quotes whenever I don't need string interpolation. The difference may be negligible but Ruby won't need to scan the string for interpolation when parsing it. I'm not going to say that I'm religious about it.

I'm 👍  on this change.

ys:
I like the fact that this will bring some stronger opinions on this framework.
I would totally go the double quotes all the things btw.
We are doing that in our projects too.

I wouldn't do the frozen header comment but that is personal.

gudmundur:
@cyberdelia When I think about this some more, I know that the single vs. double quote rule will cause headaches. For the code bases I work on, we're not consistent on this. If there were a set opinionated Pliny standard, I think it should be to always use double quotes.

```
StringLiterals:
  EnforcedStyle: double_quotes
  Severity: error
```

What do you think about that?

gudmundur:
And yeah, I agree with @ys's sentiment on `# frozen_string_literal: true`. Feels a little weird to sprinkle that all over.

ys:
Specially true that in the future it will be the default anyway.

gudmundur:
Another thought, I really do like having Rubocop on by default. Maybe the line should be drawn on checks that verify correctness, safety and readability, rather than personal styles.

cyberdelia:
I have no strong feeling about the rules we enable/disable put in ``.rubocop.yml``.

brandur:
@cyberdelia Yo! It looks to me like people are generally very positive on this one with some very minor grievances. Would you be willing to maybe revert the single/double quote rule and changes and possibly remove the `# frozen_string_literal: true` directives? I think with those two changes in we can merge and this and start moving forward.

I noticed a bunch of style discrepancies (usually around curly brace styling) just in the Pliny core code while making a change recently and I'd love to get some enforcement on this.

brandur:
Also, wow, look at GitHub's cool new web-based branch conflict resolver!! Very cool stuff.

cyberdelia:
@brandur The quote cop now promotes double-quote usage, and I reverted the frozen comment changes too.

This only apply to the template though, so you might want add rubocop rules for Pliny itself too.

brandur:
> @brandur The quote cop now promotes double-quote usage, and I reverted the frozen comment changes too.

Thanks! @gudmundur, what say you, go for the merge here?

> This only apply to the template though, so you might want add rubocop rules for Pliny itself too.

Yeah, +1. The project itself could definitely use some standards.

gudmundur:
👍

PR #286, opened 2016-12-01
Pluggable metrics backend

appleton:
This PR abstracts the metrics interface from the implementation allowing for different metric collection strategies. It also adds a new metrics backend for direct to Librato logging.

- [x] Create a “Logger” backend for metrics
- [x] Use the logger backend by default in the metrics module
- [x] Allow different metrics backends to be set and unset
- [x] Implement a new Librato backend as a drop in replacement for direct to Librato metrics https://github.com/appleton/pliny-librato


appleton:
One big thing about this which I think we still need to address is that this implementation will just make all librato metric calls in-process and in the same thread. I have a feeling that's not going to work too well for us.

I was wondering if there's a way we can use `Librato::Queue` to gather metrics and then have that "submit" to a sidekiq job every so often for reporting.

appleton:
TODO:
- Change to allow multiple metrics backends
- Configure default logger backend in an initializer
- Update librato backend to submit metrics in a new thread, maybe with sucker punch

appleton:
This should be ready for an initial review now. Still todo:

- [ ] Install in a project and see it work end to end.
- [ ] Add documentation.

lib/pliny/metrics/backends/librato.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  MSG
+
+  raise
+end
+
+module Pliny
+  module Metrics
+    module Backends
+      class Librato
```
appleton: I don't love that this is a class and the Logger backend is a module with class methods. Unfortunately we need an instance of this so that we can attach all of the `Concurrent::Async` stuff. Open to suggestions on how to make this consistent (maybe just require that `Backends::Logger` is a class too?).

lib/pliny/metrics/backends/librato.rb
```diff
@@ -0,0 +1,50 @@
...
+! Please add:
+!   gem "librato-metrics"
+!   gem "concurrent-ruby"
+! to your gemfile and run `bundle install` if you wish to use the Librato
+! metrics backend
+
+  MSG
+
+  raise
+end
```
gudmundur: I mentioned this when we discussed this, but I believe this should be in a separate gem, for example `pliny-librato`

lib/pliny/metrics/backends/librato.rb
gudmundur: Makes sense that `Backends::Logger` is a class as well. That aligns with `ErrorReporters::Rollbar`.

lib/template/config/initializers/metrics.rb
```diff
@@ -0,0 +1 @@
+Pliny::Metrics.backends << Pliny::Metrics::Backends::Logger
```
gudmundur: I'm assuming that the Librato authentication needs to happen here?

lib/pliny/metrics.rb
```diff
@@ -1,29 +1,48 @@
...
-    def self.count(*names, value: 1)
-      counts = Hash[names.map { |n| [metric_prefix(:count, n), value] }]
-      Pliny.log(counts)
+    extend self
+
+    attr_accessor :backends
+
+    @backends = []
+
+    def reset_backends
```
gudmundur: I'm assuming this is only a handy helper for tests? If so, can this be a helper method in the specs instead?

lib/pliny.rb
```diff
@@ -8,6 +8,7 @@
 require_relative "pliny/helpers/params"
 require_relative "pliny/helpers/serialize"
 require_relative "pliny/log"
+require_relative "pliny/metrics/backends/logger"
```
gudmundur: Any reason why this is required before `pliny/metrics`? Looks a little off having it in this order.

gudmundur:
@brandur I'd love to get your eyes on this as well, as we discussed this at some point as well.

appleton:
I've extracted a gem https://github.com/appleton/pliny-librato

lib/template/config/initializers/metrics.rb
appleton: No, that can happen in an initializer within the actual app.

gudmundur:
@appleton @joshwlewis This stuff is awesome! 👍  Before bringing this in, I'd love to get @brandur's eyes on this as we discussed this at some point.

joshwlewis:
> I think that if you just upgraded your app's pliny gem without getting the initializer you'd just end up with a metrics module with an empty set of backends.

Yeah, that's certainly true. I think removing the initializer and defaulting the backends to `[Pliny::Metrics::Backends::Logger]` should well for backwards compatibility, and still provide the flexibility we're looking for.

> has the l2met style logging been determined to just be too heavyweight and so it's set to be dropped in favor of a more direct interface?

Yes, exactly. We'd like to be more explicit about what we send to our metrics services. Sending the entire log stream to metrics providers is a bit heavy-handed, and doesn't comply with our latest security model.

spec/metrics_spec.rb
```diff
@@ -8,15 +8,22 @@
...
-    Pliny::Metrics.backends << test_backend
     allow(Config).to receive(:app_name).and_return('pliny')
   end
 
-  after do
-    Pliny::Metrics.backends = []
+  around do |example|
+    original_backends = Pliny::Metrics.backends
+    example.run
+    Pliny::Metrics.backends = original_backends
```
brandur: One minor suggestion: maybe wrap these in a set of `begin`/`ensure` blocks.

brandur:
> Yeah, that's certainly true. I think removing the initializer and defaulting the backends to [Pliny::Metrics::Backends::Logger] should well for backwards compatibility, and still provide the flexibility we're looking for.

Yeah, I think this is the right call (along with leaving in the initializer so that new projects get it). If we do another major version increment, we can always remove the default at that time (and replace a call to an empty set of backends with an error).

I left a minor comment on your patch, but +1.

> Yes, exactly. We'd like to be more explicit about what we send to our metrics services. Sending the entire log stream to metrics providers is a bit heavy-handed, and doesn't comply with our latest security model.

Makes sense! I'm glad to hear about progress on that front. The l2met style logging was always an interesting idea, but was very obviously wasteful once you hit any kind of non-trivial scale.

I also like how this opens a few more doors: I'd strongly champion certain services that are full stop amazing (say a Mailgun or a Rollbar), but Librato isn't one of them. At Stripe we're using DataDog, and I have to say that compares quite favorably to Librato. It's nice to know that after this patch lands, I'd easily be able to plug in a metrics driver for it.

joshwlewis:
Cool, thanks @brandur. That's good feedback all around. I've updated this PR with everything mentioned. 

> I'd easily be able to plug in a metrics driver for it.

Yep, this is working nicely for heroku/pliny-librato. Not only can we control how and where metrics are reported, but how we schedule the API calls. We're just kicking off threads right now, but we could also queue them up and report in batches.

gudmundur:
Awesome work @appleton and @joshwlewis!

PR #287, opened 2016-12-02
Don't allow to require inexistent path

cyberdelia:
This change how Pliny require files, and do not allow to look for an inexistent path.
This avoid path bloat in the app initializer.


brandur:
Makes sense to me and seems to be more consistent with what's already in there. +1.

PR #288, opened 2016-12-07
Pliny::Metrics.measure value argument

joshwlewis:
Many times, the operation you want to measure is not block execution time. You may want to measure something that is already stored as a variable. This is a common pattern for us, and we commonly do this:

```ruby
      queue = ::Sidekiq::Queue.new

      Pliny.log(
        "app.sidekiq.latency" => queue.latency,
        "app.sidekiq.depth" => queue.depth
      )
```

You can see here that we're leaning on l2met logging, but with the new pluggable metrics backends, it'd be much nicer to be able to do something like:

```ruby
Pliny::Metrics.measure("app.sidekiq.latency", queue.latency)
```

Right now, `measure` only reports execution time of the provided block. This PR proposes that the block becomes optional and if provided, the `:value` option takes precedence.

This should provide a symmetry between the two methods, e.g.:

```ruby
Pliny::Metrics.count(:foo, value: 3)
Pliny::Metrics.measure(:bar, value: 3.14)
```

But should also maintain the current API of:

```ruby
Pliny::Metrics.measure(:qux) { # some long running work }
```

### DIFF
--- a/lib/pliny/metrics.rb
+++ b/lib/pliny/metrics.rb
@@ -24,7 +24,7 @@ module Pliny
       measurement =
         if opts.has_key?(:value)
           opts[:value]
-        elsif !elapsed.nil?
+        elsif block
           elapsed
         else
           0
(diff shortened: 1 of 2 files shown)


gudmundur:
Nice work @joshwlewis!

PR #289, opened 2016-12-08
Metrics middleware

joshwlewis:
This adds a new middleware that is extracted from the operational patterns we rely on for our Ruby apps at Heroku. It sends a basic, minimal number of request metrics to the configured backend provider. These metrics should allow an operator to get a basic idea of:

1. How many requests the app is processing
2. The breakdown of statuses the app is responding with
3. The overall request latency for the app

I've added this as an independent middleware so that apps can opt out of the functionality if desired.


pedro:
very cool! :+1:

brandur:
Love it! Really nice job on the tests too. +1.

gudmundur:
I wouldn't mind seeing this as a default in `routes.rb`. If people don't want this, they can easily opt out. It also helps with discovery of this feature. Yum yum.

joshwlewis:
Hmm, it is in `routes.rb`. Shouldn't that make it the default?

gudmundur:
Lol. Totally missed that.

PR #291, opened 2016-12-09
Handle an empty rack_env in rollbar reporter

appleton:
In the `error_reporter` we default to an empty hash for `rack_env`, but if that is passed to rollbar it results in an exception like:

```
NoMethodError: "undefined method `rewind' for nil:NilClass" in /app/vendor/bundle/ruby/2.3.0/gems/rollbar-2.11.3/lib/rollbar/request_data_extractor.rb:157:in `rollbar_raw_body_params': build_item in exception_data
```

That's because we call the `#extract_request_data_from_rack` method which expects a fully formed rack env to be passed. I hit this error when trying to use the Pliny error reporter from outside of a rack request context.

This change removes the request scoping if no rack env was passed.

### DIFF
--- a/lib/pliny/error_reporters/rollbar.rb
+++ b/lib/pliny/error_reporters/rollbar.rb
@@ -18,9 +18,11 @@ module Pliny
       private
 
       def fetch_scope(context:, rack_env:)
-        {
-          request: proc { extract_request_data_from_rack(rack_env) }
-        }
+        scope = {}
+        if rack_env.respond_to?(:body)
+          scope[:request] = proc { extract_request_data_from_rack(rack_env) }
+        end
+        scope
       rescue Exception => e
         report_exception_to_rollbar(rack_env, e)
         raise
--- a/spec/error_reporters/rollbar_spec.rb
+++ b/spec/error_reporters/rollbar_spec.rb
@@ -1,0 +1,56 @@
+require "spec_helper"
+require "rollbar"
+require "pliny/error_reporters/rollbar"
+
+describe Pliny::ErrorReporters::Rollbar do
+  subject(:reporter) { described_class.new }
+
+  describe "#notify" do
+    let(:exception) { StandardError.new("Something went wrong") }
+    let(:context)   { {} }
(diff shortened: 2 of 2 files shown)


spec/error_reporters/rollbar_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    let(:rack_env)  { double(:rack_env, body: StringIO.new) }
+
+    subject(:notify) do
+      reporter.notify(exception, context: context, rack_env: rack_env)
+    end
+
+    before do
+      allow(::Rollbar).to receive(:reset_notifier!)
+      allow(::Rollbar).to receive(:scoped).and_yield
+      allow(reporter).to receive(:report_exception_to_rollbar)
```
gudmundur: I hear you. I'd prefer not breaking APIs for a test.

lib/pliny/error_reporters/rollbar.rb
```diff
@@ -18,9 +18,11 @@ def notify(exception, context:, rack_env:)
       private
 
       def fetch_scope(context:, rack_env:)
-        {
-          request: proc { extract_request_data_from_rack(rack_env) }
-        }
+        scope = {}
+        if rack_env.respond_to?(:body)
```
gudmundur: Is this the "accepted" way to detect whether a hash is indeed a `rack_env`? Wouldn't `rack_env.empty?` suffice?

appleton: It's more that this is the method that is called internally so I figure a duck type check is sufficient

spec/error_reporters/rollbar_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    context "given an empty rack_env" do
+      let(:rack_env) { {} }
+
+      it "reports to Rollbar with an empty scope" do
+        notify
+        expect(Rollbar).to have_received(:scoped).once.with({})
+      end
+
+      it "delegates to #report_exception_to_rollbar" do
+
```
appleton: Umm, no. No it isn't :D

gudmundur:
👍

PR #292, opened 2016-12-11
Deprecation note on ConfigHelpers

brandur:
Adds a small deprecation note to ConfigHelpers to make it clear that CastingConfigHelpers is the way forward here.

@gudmundur Mind doing a basic review/sanity check on this one?


gudmundur:
👍

PR #293, opened 2016-12-12
Add canonical log line middleware to Pliny

brandur:
A few weeks ago I wrote about the idea of canonical log lines [1], a technique that we use inside of Stripe. The basic idea is to build one big log line that emits a whole bunch of vitals on an API request all at once for easy reference. It's possible to get similar information by using something like Splunk's `transaction` command and tie a bunch of log lines together by request ID, but this is slower and fairly inconvenient.

Subjectively, we find these incredibly useful internally. I'd go as far as to say that _most of the time_ when referencing Splunk information, people don't even bother to look at any log traces produced by our API except for the canonical log line.

This just emits a line to stdout for probably ingestion into Splunk, but bigger users might want to consider emitting this elsewhere for long-term archiving. Putting them onto a Kafka topic (probably async for performance reasons and because some losiness is acceptable) that feeds into Redshift, Postgres (especially with the upcoming built-in table partitioning), Citus, or some other data warehouse. We send ours over NSQ into Redshift which allows us to build profiles for the last three months worth of requests. Lower volume APIs could conceivably store them for much longer.

I've replaced the `Instruments` middleware with the canonical log line middleware because they're performing a nearly identical function. `Instruments` is still available for backwards compatibility.

@gudmundur @pedro (and anyone else who compares to comment) Thoughts on bringing this mainline?

[1] https://brandur.org/canonical-log-lines


pedro:
Nice! Been meaning to try canonical logs at Lugg so this should be really helpful.

It looks good to me! Just two questions here:

- Why the types on log fields? Is that to help storing the canonical line in a more structured store? Or just to validate that you're logging in the expected format?

- How would you go about adding id/email of the authenticated user, for instance? Is the idea that you save that stuff in the rack env then fetch it from the middleware?

brandur:
Thanks Pedro!

> Why the types on log fields? Is that to help storing the canonical line in a more structured store? Or just to validate that you're logging in the expected format?

This are purely my desire for more type safety in Ruby showing through and are not strictly necessary. Like you suggest, their major feature is that they help us be more certain that we're not storing anything into a field that we didn't expect. Another one is that they help a human browsing the code to see more easily what the field should look like.

> How would you go about adding id/email of the authenticated user, for instance? Is the idea that you save that stuff in the rack env then fetch it from the middleware?

Yeah! The idea is basically that communication should happen through a request's env.

Sinatra defines a certain interface by using keys of the format `sinatra.*` like `sinata.error`. Here I introduce a similar interface for Pliny that looks like `pliny.*` such as `pliny.error` (and other keys like `REQUEST_ID` should probably be eventually moved to `pliny.request_id`). I'd further postulate that apps should use their own keys at `my_app.*` like `my_app.user`. This is how we do it at Stripe.

I think one weakness of this approach is that the entire env contract is not explicitly defined anywhere, and middleware and other modules are allowed to just drop in new keys wherever, which can lead to an obscure and brittle interface. I still don't have a great answer for that, but in our equivalent of a `routes.rb` I try to hint at where certain critical keys start to come into play. For example:

``` ruby
...

use Middleware::Authenticator

#
# `stripe.key` available in env below this line
#

use Middleware::UserLoader

#
# `stripe.user` available in env below this line
#

run App
```

The interfaces are still far too implicit for my taste, but it mostly works.

lib/pliny/middleware/canonical_log_line.rb
```diff
@@ -0,0 +1,121 @@
...
+  # be:
+  #
+  #     * ID and email of an authenticated user.
+  #     * ID of API key used, OAuth application and scope.
+  #     * Remaining and total rate limits.
+  #     * Name of the service, HEAD revision, release number.
+  #     * Name of the internal system that initiated the request.
+  #
+  class CanonicalLogLine
+    class CanonicalLogLine
```
gudmundur: This took me a second to understand, got confused here. We've got a nested class inside the middleware with the same name. I'm assuming that the inner one is a value object to make constructing the canonical log line easier?

lib/pliny/middleware/canonical_log_line.rb
```diff
@@ -0,0 +1,121 @@
+module Pliny::Middleware
+  # Emits a "canonical log line", i.e. a single log line that contains as much
+  # relevant information about a request as possible and which makes for a
+  # single convenient reference point to understand all the vitals of any
+  # single request.
+  #
+  # This default implementation contains some useful data to get a project
+  # started, but it's usually recommended to vendor this middleware into your
+  # project and start adding some custom fields. Some examples of those might
+  # be:
```
gudmundur: If we're recommending that this middleware is vendored in, does it make sense to include it in the template instead?

lib/pliny/middleware/rescue_errors.rb
```diff
@@ -6,14 +6,20 @@ def initialize(app, options = {})
...
       Pliny::Errors::Error.render(e)
     rescue => e
+      error = e
       raise if @raise
 
       Pliny::ErrorReporters.notify(e, rack_env: env)
       Pliny::Errors::Error.render(Pliny::Errors::InternalServerError.new)
+    ensure
+      # Leave this in env for CanonicalLogLine.
+      env["pliny.error"] = e
```
gudmundur: Can't we just duplicate this line in each `rescue` block? I'm assuming you're explicitly setting it in this way to illustrate the point that it should always be set. If you'd like to retain that explicitness, how about using an `attr_reader :error`? The `error = nil` in the main body weirds me out a little.

lib/pliny/middleware/canonical_log_line.rb
```diff
@@ -0,0 +1,121 @@
...
+          #
+
+          line.timing_total_elapsed = (Time.now - start).to_f
+
+          @emitter.call(line.to_h)
+        rescue => e
+          # We hope that a canonical log line never fails, but in case it
+          # does, do not fail the request because it did.
+          Pliny.log(message: "Failed to emit canonical log line")
+          Pliny.log_exception(e)
```
gudmundur: Can you pass this to `Pliny::ErrorReporters#notify` as well? Then it'll show up in Rollbar and friends.

lib/pliny/middleware/canonical_log_line.rb
brandur: Good question!

I definitely considered doing so, but decided against it (for now) for two reasons:

1. I think there's a pretty good likelihood that this isn't going to be the first thing newer projects want to start modifying. The default is probably going to serve you pretty well until you're in production and want to start building up some observability.
2. Because there's a chance that my "from scratch" implementation is not bug free, it would be nice to give this a little time to sit in master so that any bug fixes can be shipped directly as a Pliny patch.

I think given a little bit of time and maybe with some vetting at some non-Stripe companies, it would totally make sense to move this over to the template.

lib/pliny/middleware/canonical_log_line.rb
brandur: Hah yeah, understandably.

I had the middleware called `CanonicalLogLineEmitter` until I renamed it at the last second to just `CanonicalLogLine` because it seemed to fit better into our existing middleware naming scheme.

I renamed the inner model to `LogLine` and annotated it with a comment in 49a084c. Hopefully this will help a little bit.

lib/pliny/middleware/canonical_log_line.rb
brandur: Nice catch. Fixed in f5abf0a.

lib/pliny/middleware/rescue_errors.rb
brandur: Yeah, good point, it's a little messy. I refactored a bit in 2e384c3.

I'm a little hesitant to use an `attr_reader` because a middleware's exposed interface `call(env)` is pretty suggestive that the instance itself should be stateless throughout any single request (i.e. you could potentially run two requests through one instance at the same time).

brandur:
Thanks for the quick review @gudmundur!

> I'm not familiar enough with the canonical log lines concept to see it's true value, but I'm sure once I start using this it'll become apparent. Or when I've slept on this.

Yeah, to be honest, I would be fairly skeptical too. I think you really have to see them in action to really start appreciating their value. [I put together an "examples" section](https://brandur.org/canonical-log-lines#examples) to try and demonstrate some of the ad-hoc query properties that they might be good for, but they probably more rightly need a full screencast or something.

In terms of moving forward here: does it seem mostly peaceful to bring in (pending another review pass on the latest changes of course)?

gudmundur:
@brandur Yeah, I'd like to see this in Pliny and I'm excited to give it a spin. I'll have another pass at this tomorrow, it's late here. 😴 .

brandur:
> @brandur Yeah, I'd like to see this in Pliny and I'm excited to give it a spin. I'll have another pass at this tomorrow, it's late here. 😴 .

No rush! Thanks.

lib/pliny/middleware/canonical_log_line.rb
gudmundur: @brandur I know for sure that when this lands, I will want to add other fields to the canonical log line. For example, on the GitHub integration, I would want to include the repository.

I do understand your reservations about bringing this into the template right away. But as I understand it, moving it to the template after a while, will also require a migration in user code. I.e. the namespace of the middleware will change.

Do you think we could do this in a way where the default lives in Pliny-land, but extensions live in user-land? It probably either means to add a context of sorts to the request, or allowing subclassing of the base type which can than be loaded.

brandur: @gudmundur Cool, yeah I would be too opposed to this.

What would you think about the `LogLine` class living in user-land but including a default module from Pliny, like:

``` ruby
class LogLine
  include Pliny::CanonicalLogLineHelpers
  include Pliny::CanonicalLogLineDefaultFields

  ...
```

The vendored middleware would then contain something like this:

``` ruby
def call(env)
  begin
    start = Time.now
    status, headers, response = @app.call(env)
  ensure
    begin
      line = LogLine.new
      Pliny::CanonicalLogLineDefaultFields.extract(line, status, headers, response,
        start: start)
      ...
```

That way we could still tweak the default fields while allowing some extensibility right away.

The other plausible option is just to vendor it completely. I think you could make the argument for that because it just makes things so much simpler and easier to look at, but at the cost of possibly missing out on bug fixes and updates.

gudmundur: @brandur I like the idea of the includes in the `LogLine` class. I'm not too keen on how messy the middleware becomes though.

Back to your earlier comment, bringing this in as-is, weed out bugs if present, then vendor it into the template to allow extensibility. I'm starting to think that's probably the best way forward. We could add it to the template by duplicating it, and adding deprecation warnings to the `Pliny::Middleware::CanonicalLogLine` with instructions on how to migrate. How does that sound?

brandur: @gudmundur Sorry for the delayed reply. Yep, that sounds amazing. Going to bring this mainline then.

lib/pliny/log.rb
```diff
@@ -4,6 +4,10 @@ def log(data, &block)
       log_to_stream(stdout || $stdout, merge_log_contexts(data), &block)
     end
 
+    def log_without_context(data, &block)
```
gudmundur: @brandur After trying this out, I realized that having at least the `default_context` makes sense. Otherwise it might be hard to search for log base on a single app: `app=my-pliny-app deployment=production canonical-log-line ...`

brandur: @gudmundur Yeah, actually that makes perfect sense in retrospect. As a slight alternative, what would you say about a setup where you dependency injected a Pliny configuration class a little like this?

``` patch
diff --git a/lib/pliny/middleware/canonical_log_line.rb b/lib/pliny/middleware/canonical_log_line.rb
index 90be252..8083ce2 100644
--- a/lib/pliny/middleware/canonical_log_line.rb
+++ b/lib/pliny/middleware/canonical_log_line.rb
@@ -24,6 +24,13 @@ module Pliny::Middleware
       include Pliny::CanonicalLogLineHelpers
 
       #
+      # general config
+      #
+
+      log_field :app_name, String
+      log_field :deployment, String
+
+      #
       # error
       #
 
@@ -56,8 +63,14 @@ module Pliny::Middleware
       log_field :timing_total_elapsed, Float
     end
 
-    def initialize(app, emitter:)
+    def initialize(app, config:, emitter:)
       @app = app
+
+      # Note that we dependency inject "config" because a global `Config`
+      # symbol is only available in the context of an actual Pliny app, and not
+      # within the test suite of Pliny itself.
+      @config = config
+
       @emitter = emitter
     end
 
@@ -70,6 +83,15 @@ module Pliny::Middleware
           line = LogLine.new
 
           #
+          # general config
+          #
+
+          if @config
+            line.app_name = @config.app_name
+            line.deployment = @config.deployment
+          end
+
+          #
           # error
           #
 
diff --git a/lib/template/lib/routes.rb b/lib/template/lib/routes.rb
index 6a48bed..531e806 100644
--- a/lib/template/lib/routes.rb
+++ b/lib/template/lib/routes.rb
@@ -5,6 +5,7 @@ Routes = Rack::Builder.new do
   use Pliny::Middleware::RequestStore::Seed, store: Pliny::RequestStore
   use Pliny::Middleware::Metrics
   use Pliny::Middleware::CanonicalLogLine,
+      config: Config,
       emitter: -> (data) {
         Pliny.log_without_context({ canonical_log_line: true }.merge(data))
       }
diff --git a/spec/middleware/canonical_log_line_spec.rb b/spec/middleware/canonical_log_line_spec.rb
index 51888d9..efab16a 100644
--- a/spec/middleware/canonical_log_line_spec.rb
+++ b/spec/middleware/canonical_log_line_spec.rb
@@ -1,12 +1,23 @@
 require "spec_helper"
 
 describe Pliny::Middleware::CanonicalLogLine do
+  class TestConfig
+    def app_name
+      "my-app"
+    end
+
+    def deployment
+      "production"
+    end
+  end
+
   def app
     Rack::Builder.new do
       run Sinatra.new {
         use Pliny::Middleware::RequestID
 
         use Pliny::Middleware::CanonicalLogLine,
+          config: TestConfig.new,
           emitter: -> (data) { Pliny.log_without_context(data) }
 
         use Pliny::Middleware::RescueErrors, raise: false
@@ -34,6 +45,10 @@ describe Pliny::Middleware::CanonicalLogLine do
     header "User-Agent", "rack-test"
     get "/apps/123"
 
+    config = TestConfig.new
+    assert_equal config.app_name, data[:app_name]
+    assert_equal config.deployment, data[:deployment]
+
     assert_match Pliny::Middleware::RequestID::UUID_PATTERN,
       data[:request_id]
     assert_equal "127.0.0.1", data[:request_ip]
```

I sort of like it because all the fields to be included in the canonical line stay very explicit. The downside is that once again it makes customization somewhat difficult until we start vendoring the middleware into new Pliny apps. Thoughts?

gudmundur: I can see arguments for both approaches. I get the motivation for keeping things explicit here. However, I do wonder if there's ever a log line that you'd want to emit without the `default_context`? Also, passing the entire `config` object into something that holds log lines seems like it could potentially leak creds.

brandur: @gudmundur Yeah, thinking about it a little more, I agree with your original assessment. I added a change in #307.

PR #294, opened 2016-12-12
Add "pry-byebug" to Pliny

brandur:
Adds "pry-byebug" to the main Pliny project as a development dependency.
Pry can be invoked already, but it's not a whole bunch of use without a
debugger.


brandur:
Hopefully this one will not be too contentious. Pulling it in.

gudmundur:
Nice!

PR #295, opened 2016-12-20
Fix rollbar scope extraction

mikehale:
A rack env is a simple hash and as such will never respond to the method `body`. It does however have a key of `rack.input`. Maybe there is some confusion around what is being passed into `notify`... `env` vs `Rack::Request.new(env)`?

/cc @appleton 
ref https://github.com/interagent/pliny/pull/291

### DIFF
--- a/spec/error_reporters/rollbar_spec.rb
+++ b/spec/error_reporters/rollbar_spec.rb
@@ -41,6 +41,10 @@ describe Pliny::ErrorReporters::Rollbar do
     context "given an empty rack_env" do
       let(:rack_env) { {} }
 
+      it "expects rack_env to be a hash" do
+        assert_kind_of(Hash, rack_env)
+      end
+
       it "reports to Rollbar with an empty scope" do
         notify
         expect(Rollbar).to have_received(:scoped).once.with({})
--- a/spec/error_reporters/rollbar_spec.rb
+++ b/spec/error_reporters/rollbar_spec.rb
@@ -8,7 +8,7 @@ describe Pliny::ErrorReporters::Rollbar do
   describe "#notify" do
     let(:exception) { StandardError.new("Something went wrong") }
     let(:context)   { {} }
-    let(:rack_env)  { double(:rack_env, body: StringIO.new) }
+    let(:rack_env)  { { "rack.input" => StringIO.new } }
 
     subject(:notify) do
       reporter.notify(exception, context: context, rack_env: rack_env)
--- a/lib/pliny/error_reporters/rollbar.rb
+++ b/lib/pliny/error_reporters/rollbar.rb
@@ -19,7 +19,7 @@ module Pliny
 
       def fetch_scope(context:, rack_env:)
         scope = {}
-        if rack_env.respond_to?(:body)
+        if rack_env.has_key?("rack.input")
           scope[:request] = proc { extract_request_data_from_rack(rack_env) }
         end
         scope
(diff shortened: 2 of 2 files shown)


brandur:
Nice find @mikehale :)

mikehale:
bump. anything from blocking this from being merged?

gudmundur:
@mikehale sorry for the delay. Is there any reason why we simply don't check for a `.is_a(Hash)`? Is the Rollbar gem expecting `rack.input` to be present?

mikehale:
> Is the Rollbar gem expecting `rack.input` to be present?

Yes, `extract_request_data_from_rack` expects `rack.input` to be an IO.

I tested reporting an exception with a `nil` `rack_env`, but the rollbar gem does not handle that either:

```
irb(main):016:0> report_exception_to_rollbar(nil, RuntimeError.new)
D, [2017-01-04T11:47:57.596340 #42833] DEBUG -- : [Rollbar] Reporting exception: RuntimeError
I, [2017-01-04T11:47:57.596896 #42833]  INFO -- : [Rollbar] Scheduling item
I, [2017-01-04T11:47:57.596926 #42833]  INFO -- : [Rollbar] Sending item
I, [2017-01-04T11:47:58.073781 #42833]  INFO -- : [Rollbar] Success
I, [2017-01-04T11:47:58.073930 #42833]  INFO -- : [Rollbar] Details: https://rollbar.com/instance/uuid?uuid=b068d0f6-fe22-4ab9-9454-cd16b08eaf49 (only available if report was successful)
W, [2017-01-04T11:47:58.075205 #42833]  WARN -- : [Rollbar] Exception while reporting exception to Rollbar: undefined method `[]=' for nil:NilClass
=> true
irb(main):017:0> report_exception_to_rollbar({}, RuntimeError.new)
D, [2017-01-04T11:48:12.218743 #42833] DEBUG -- : [Rollbar] Reporting exception: RuntimeError
I, [2017-01-04T11:48:12.219101 #42833]  INFO -- : [Rollbar] Scheduling item
I, [2017-01-04T11:48:12.219126 #42833]  INFO -- : [Rollbar] Sending item
I, [2017-01-04T11:48:12.682914 #42833]  INFO -- : [Rollbar] Success
I, [2017-01-04T11:48:12.682982 #42833]  INFO -- : [Rollbar] Details: https://rollbar.com/instance/uuid?uuid=e6bcc9a0-c688-4075-87b7-b55839d5e296 (only available if report was successful)
D, [2017-01-04T11:48:12.683011 #42833] DEBUG -- : [Rollbar] Exception uuid saved in env: e6bcc9a0-c688-4075-87b7-b55839d5e296
```

I could change the check to `rack_env.empty?` which would be more generic, not tied to the rollbar interface as much, but still should work in the case where a `rack_env` is not actually populated, which I think was the intent of #291.

spec/error_reporters/rollbar_spec.rb
```diff
@@ -41,6 +41,10 @@
     context "given an empty rack_env" do
       let(:rack_env) { {} }
 
+      it "expects rack_env to be a hash" do
+        assert_kind_of(Hash, rack_env)
```
appleton: Is this test needed? It looks like it is just asserting that L42 is a Hash which seems redundant.

appleton:
LGTM 👍

gudmundur:
👍

spec/error_reporters/rollbar_spec.rb
mikehale: it is somewhat redundant, but it is also trying to be extra clear so that the test double is not replaced by some other type of object

mikehale:
thanks!

gudmundur:
@mikehale thank you! 🙂

PR #297, opened 2017-01-06
Upgrade Puma to 3.x series

fdr:
Why not maybe. I've been doing this in prod for a while and use of `pliny-upgrade` has shown me that the template still uses a 2.x release.


brandur:
Looks good to me. Thanks!

PR #300, opened 2017-01-09
Fix deprecated mock of #exists?

gudmundur:
A [recent change correcting Rubocop violations](https://github.com/interagent/pliny/pull/285) changed `File.exists?` to `File.exist?`, but a spec mocking out the deprecated method was not updated. This change addresses that failing spec

cc: @cyberdelia @brandur


brandur:
Looks good to me! Thanks @gudmundur.

PR #302, opened 2017-01-10
Make pliny compatible with ruby 2.4

cyberdelia:
This allow to run Pliny with ruby 2.4 by upgrading the ``activesupport`` gem, since it requires a version of the ``json`` gem that is incompatible with ruby 2.4.

Sadly this also means we have to drop support for ruby > 2.2.2 since ``activesupport`` do not support them.


brandur:
@cyberdelia @gudmundur Do you know if the loss of 2.2.2 support is going to be a problem for anyone? It seems like Heroku's generally doing a pretty good job of keeping its Ruby stacks up-to-date.

brandur:
@cyberdelia Mind fixing that one canonical line test this way instead?

``` patch
diff --git a/spec/canonical_log_line_helpers_spec.rb b/spec/canonical_log_line_helpers_spec.rb
index 3509912..0290dec 100644
--- a/spec/canonical_log_line_helpers_spec.rb
+++ b/spec/canonical_log_line_helpers_spec.rb
@@ -22,9 +22,9 @@ describe Pliny::CanonicalLogLineHelpers do
   it "rejects values that are of the wrong type" do
     line = TestCanonicalLogLine.new
     e = assert_raises ArgumentError do
-      line.field_string = 42
+      line.field_string = true
     end
-    assert_equal "Expected field_string to be type String (was Fixnum)",
+    assert_equal "Expected field_string to be type String (was TrueClass)",
       e.message
   end
```

I think it's a little less janky because it's version agnostic. Thanks!

cyberdelia:
@brandur I did just that.

brandur:
Sweet! Well, as long as we're reasonably sure we're not going to leave too many people around by dropping 2.1, LGTM.

gudmundur:
Did a quick search through the Heroku repos and most Ruby projects seem to be 2.3+. I'm bringing this in.

PR #303, opened 2017-02-06
Include custom data in scope sent to rollbar

mikehale:
This includes the provided context in the rollbar scope. Previously the context was ignored and not included in the scope. This is similar to how scope is defined for rails in rollbar: https://rollbar.com/docs/notifier/rollbar-gem/#the-scope

### DIFF
--- a/lib/pliny/error_reporters/rollbar.rb
+++ b/lib/pliny/error_reporters/rollbar.rb
@@ -18,7 +18,7 @@ module Pliny
       private
 
       def fetch_scope(context:, rack_env:)
-        scope = {}
+        scope = { custom: context }
         unless rack_env.empty?
           scope[:request] = proc { extract_request_data_from_rack(rack_env) }
         end
--- a/spec/error_reporters/rollbar_spec.rb
+++ b/spec/error_reporters/rollbar_spec.rb
@@ -7,7 +7,7 @@ describe Pliny::ErrorReporters::Rollbar do
 
   describe "#notify" do
     let(:exception) { StandardError.new("Something went wrong") }
-    let(:context)   { {} }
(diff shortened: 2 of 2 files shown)


mikehale:
Hmm, this does not seem to be working correctly. I'll investigate and update when this is ready.

mikehale:
@gudmundur this is ready for review now.

.travis.yml
```diff
@@ -9,6 +9,8 @@ before_install:
   - gem update --system
 before_script:
   - createdb pliny-gem-test
+before_install:
+  - gem update --system
```
gudmundur: Would you mind removing this? This has already been addressed with https://github.com/interagent/pliny/commit/ae7dc845184267b81af7b5de56bcb8b42fe56596.

lib/pliny/error_reporters/rollbar.rb
```diff
@@ -18,7 +18,7 @@ def notify(exception, context:, rack_env:)
       private
 
       def fetch_scope(context:, rack_env:)
-        scope = {}
+        scope = { custom: context }
```
gudmundur: Weakly held opinion, but is there a reason we don't call this `context`? Does that name hold some significance to Rollbar?

mikehale: It does hold significance to rollbar. Outside of the request object, they require data be under a `custom` key. See https://rollbar.com/docs/api/items_post/.

mikehale:
@gudmundur comments addressed.

gudmundur:
@mikehale Cheers! 🍻

mikehale:
Thanks!

PR #305, opened 2017-02-15
Fix Travis and canonical log lines

gudmundur:
While cutting a new release of Pliny, I noticed that the last two PRs we've merged introduced errors.

* The rainbow gem fails to build with the system version of RubyGems.
* A missing `)` in `routes.rb` causes failure to start.


brandur:
Fuuuu! Looks good to me Gudmundur. Thanks for the fixes!

PR #306, opened 2017-02-19
Add arity and timings for serializers to canonical log lines

gudmundur:
This change adds `serialization_arity` and `timing_serialization` to the canonical log line. I find it a little strange storing this in `env[]`, but don't really have better suggestions.

### DIFF
--- a/lib/pliny/helpers/serialize.rb
+++ b/lib/pliny/helpers/serialize.rb
@@ -14,11 +14,11 @@ No serializer has been specified for this endpoint. Please specify one with
         eos
       end
 
-      env['serialization_arity'] = data.respond_to?(:size) ? data.size : 1
+      env['pliny.serialization_arity'] = data.respond_to?(:size) ? data.size : 1
 
       start = Time.now
       serializer_class.new(structure).serialize(data).tap do
-        env['timing_serialization'] = (Time.now - start).to_f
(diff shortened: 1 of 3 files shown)


lib/pliny/helpers/serialize.rb
```diff
@@ -5,14 +5,21 @@ def self.included(base)
...
+
+      if serializer_class.nil?
         raise <<-eos.strip
 No serializer has been specified for this endpoint. Please specify one with
 `serializer Serializers::ModelName` in the endpoint.
         eos
       end
 
-      self.class.serializer_class.new(structure).serialize(data)
+      env['serialization_arity'] = data.respond_to?(:size) ? data.size : 1
```
brandur: Can we do `pliny.serialization_arity` for good env hygiene?

lib/pliny/helpers/serialize.rb
```diff
@@ -5,14 +5,21 @@ def self.included(base)
...
+
+      start = Time.now
+      serializer_class.new(structure).serialize(data).tap do
+        env['timing_serialization'] = (Time.now - start).to_f
```
brandur: And maybe `pliny.serialization_timing` here. We can flip the words back around when adding it to the log line.

gudmundur:
@brandur updated.

brandur:
Thanks! Looks good to me.

PR #307, opened 2017-02-21
By default, log canonical lines with just default context

brandur:
In retrospect, it seems to make sense to log canonical lines with just
default logging context so that it'll include an app name and any other
information that a user has deemed critical. There is some trade off
here because the information included by the canonical logger becomes
less explicit, but that'll probably be okay.

cc @gudmundur Would you mind taking a look at this one? Thanks!


gudmundur:
👍

PR #309, opened 2017-04-04
Allow an injectable scrubber Proc

elight:


elight:
We're trusting the `scrubber` to do the right thing and not overwrite existing data.

lib/pliny/log.rb
```diff
@@ -92,6 +100,7 @@ def log_context
 
     def log_to_stream(stream, data, &block)
       unless block
+        data = scrubber.call(data) if scrubber
```
gudmundur: Can you maybe check that `scrubber` is a proc or responds to `call`?

lib/pliny/log.rb
```diff
@@ -72,6 +72,14 @@ def stderr
       @stderr
     end
 
+    def scrubber=(s)
```
gudmundur: What do you think about calling this `log_scrubber` instead? From the outside it provides more context than `Pliny.scrubber`.

elight:
Sounds reasonable

lib/pliny/log.rb
elight: Ah, right. I thought about it in terms of `Pliny::Log` but, yes, because `Pliny` gets extended with `Pliny::Log`, yes? Good point.

lib/pliny/log.rb
elight: Indeed

elight:
I feel silly but I couldn't get `expect { Pliny.log_scrubber = Object.new }.to raise_error(ArgumentError)` to behave itself so I wrote my own little inline ugly version that works.

lib/pliny/log.rb
gudmundur: I'm sorry I missed this yesterday, but can you rename the `s` parameter to `scrubber`?

lib/pliny/log.rb
```diff
@@ -72,6 +72,17 @@ def stderr
       @stderr
     end
 
+    def log_scrubber=(s)
```
gudmundur: I'm sorry I missed this yesterday, but can you rename the `s` parameter to `scrubber`?

gudmundur:
I'm assuming the intention is that users of this API will specify the scrubber in `config/initializers/logging.rb`?

elight:
Sure. And yes  in their initializer.

elight:
Git issues. It's updated on the PR now.

PR #311, opened 2017-05-18
Remove hipchat notification

dmathieu:
I believe this hipchat notification is from an internal Heroku channel.
Heroku isn't using hipchat anymore, so this probably goes nowhere.


brandur:
Yep, look good to me!

PR #314, opened 2017-08-20
Add rollback for integer migrations

luax:
Makes `db:rollback` work when using "integer migrations". The table `schema_info` is used for integer migrations and `schema_migrations` for timestamped migrations (as defined in http://sequel.jeremyevans.net/rdoc/classes/Sequel/Migrator.html).

### DIFF
--- a/lib/pliny/db_support.rb
+++ b/lib/pliny/db_support.rb
@@ -50,16 +50,25 @@ module Pliny
     end
 
     def rollback
-      return unless db.tables.include?(:schema_migrations)
-      return unless current = db[:schema_migrations].order(Sequel.desc(:filename)).first
+      target = timestamp_migrations_target if db.tables.include?(:schema_migrations)
+      target = integer_migrations_target if db.tables.include?(:schema_info)
+      Sequel::Migrator.apply(db, "./db/migrate", target) if target
+    end
 
+    def timestamp_migrations_target
+      return unless current = db[:schema_migrations].order(Sequel.desc(:filename)).first
       migrations = Dir["./db/migrate/*.rb"].map { |f| File.basename(f).to_i }.sort
-      target     = 0 # by default, rollback everything
-      index      = migrations.index(current[:filename].to_i)
-      if index > 0
-        target = migrations[index - 1]
-      end
-      Sequel::Migrator.apply(db, "./db/migrate", target)
+      target = 0 # by default, rollback everything
+      index = migrations.index(current[:filename].to_i)
+      target = migrations[index - 1] if index > 0
+      target
+    end
+
+    def integer_migrations_target
+      return unless current = db[:schema_info].first
+      version = current[:version]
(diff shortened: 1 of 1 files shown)


brandur:
Looks sane to me. Thanks Ludvig!

One possible ideological doubt is that given Pliny's philosophy of strong opinions, we should probably be trying to heard users towards just one type of migration. That said, this seems like it's unlikely to cause much divergence because it only looks at what exists in the database already.

Thoughts @gudmundur?

luax:
> One possible ideological doubt is that given Pliny's philosophy of strong opinions, we should probably be trying to heard users towards just one type of migration.

Ah I understand! In this case timestamps were not used prior to Pliny and everything but rollback seemingly works. I guess I could switch to use timestamps instead :)

gudmundur:
The `pliny-generate migration hello` tooling would also create the migration files using timestamps, so there would be an inconsistency in the naming schemas.

Personally I'd be inclined to say that we should only support one or the other. In your case though, I see where that might become problematic given that the database has already been migrated using the other format. What I'd be curious to hear is how the rollback fails, does it pick the wrong migration, does it blow up, or what?

luax:
> Personally I'd be inclined to say that we should only support one or the other.

Yeah, I thought it was just missing but now that I have looked at the code more closely it seems that way. 

> What I'd be curious to hear is how the rollback fails, does it pick the wrong migration, does it blow up, or what?

Nothing happens in my case it just doesn't find the `schema_migrations` table and do nothing (which is good).

I think I will solve this by just overriding the `rollback` rake task instead, I think that would work and I could still use `Pliny::DbSupport` etc.

Thanks!

gudmundur:
@luax Great to hear. Am I understanding you correctly that it's fine to close down this PR?

luax:
Yes!

This is what I ended up with in case anyone want to do the same:

File `tasks/db.rake` overrides `rollback`:
````
require "logger"
require "uri"
require "pliny/utils"

begin
  require "sequel"
  require "sequel/extensions/migration"
  require "pliny/db_support"

  namespace :db do
    desc "Rollback last database migration"
    task :rollback do
      next if Dir["./db/migrate/*.rb"].empty?
      database_urls.each do |database_url|
        Pliny::DbSupport.run(database_url, $stdout) do |helper|
          db = helper.instance_variable_get(:@db)
          target = integer_migrations_target(db) if db.tables.include?(:schema_info)
          Sequel::Migrator.apply(db, "./db/migrate", target) if target
          puts "Rolled back `#{name_from_uri(database_url)} to `#{target}`"
        end
      end
    end

    private

    def integer_migrations_target(db)
      current = db[:schema_info].first!
      version = current[:version]
      version > 0 ? version - 1 : 0
    end
  end
rescue LoadError
  puts "Couldn't load sequel. Skipping database tasks"
end
```

gudmundur:
Thanks for your help @luax

PR #317, opened 2017-11-08
Use IndifferentHash class instead of method

dmathieu:
The indifferent_params method doesn't exist anymore in Sinatra 2.0 (See https://github.com/sinatra/sinatra/commit/44c3cb89b7c9e26ce66e0e9bfe5a481fbef6fc22#diff-f2bc16b264ca20ca842cf15d82960d34 ).

cc @bigkevmcd

### DIFF
--- a/lib/pliny/helpers/params.rb
+++ b/lib/pliny/helpers/params.rb
@@ -8,13 +8,22 @@ module Pliny::Helpers
 
     def parse_body_params
       if request.media_type == "application/json"
-        p = Sinatra::IndifferentHash[MultiJson.decode(request.body.read)]
+        p = load_params(MultiJson.decode(request.body.read))
         request.body.rewind
         p
       elsif request.form_data?
-        Sinatra::IndifferentHash[request.POST]
+        load_params(request.POST)
       else
         {}
+      end
+    end
+
+    def load_params(data)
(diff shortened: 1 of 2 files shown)


brandur:
Thanks for the tests! This looks good to me (also, this makes way more sense as a class method so I'm relatively glad to see that Sinatra changed it even if it's annoying).

Do you know if @gudmundur or someone else can run a release? I haven't done one in a while.

gudmundur:
Thanks @dmathieu. Will this work with Sinatra 1? I know the class exists, but I'm not sure it'll behave the same way.

If this doesn't work with Sinatra 1, we can either:

1. Add a support method that'll detect whether the method or class should be used. Thus support both versions.
2. Drop support for Sinatra 1, and cut a major version of Pliny.

My personal preference would be to support both, with a comment about dropping some code in the future once we go Sinatra 2 only.

dmathieu:
Very good point. I've just added support for both.

gudmundur:
@dmathieu Nice work. This looks good to me. @brandur does this seem solid to you as well?

brandur:
Yeah, looks good to me! Thanks @dmathieu, and nice catch on that one @gudmundur.

PR #318, opened 2017-11-21
Run tests using sinatra 1 as well as sinatra 2

dmathieu:


.travis.yml
```diff
@@ -3,6 +3,9 @@ rvm:
   - 2.4.0
   - 2.3.3
   - 2.2.2
+gemfile:
+  - Gemfile
+  - Gemfile.sinatra1
```
gudmundur: Do you mind renaming the file to `Gemfile.sinatra-1.x`? It's a minor change, but it feels clearer on what it's doing.

PR #320, opened 2017-11-30
Add mechanism for additional headers.

bigkevmcd:
To add additional headers to responses, we have to resort to something like...

```Ruby
allowed_cors_headers = Pliny::Middleware::CORS.send(:remove_const, :ALLOW_HEADERS).dup
allowed_cors_headers << "X-Origin"

Pliny::Middleware::CORS::ALLOW_HEADERS = allowed_cors_headers.freeze
```

It'd be nice if we could just configure the response headers globally when configuring our app?

### DIFF
--- a/lib/pliny/middleware/cors.rb
+++ b/lib/pliny/middleware/cors.rb
@@ -8,6 +8,12 @@ module Pliny::Middleware
     EXPOSE_HEADERS =
       %w( Cache-Control Content-Language Content-Type Expires Last-Modified Pragma ).freeze
 
+    @@additional_headers = []
+
+    def self.add_additional_header(header)
+      @@additional_headers << header
+    end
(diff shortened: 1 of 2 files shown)


lib/pliny/middleware/cors.rb
```diff
@@ -32,11 +38,15 @@ def cors_request?(env)
       env.has_key?("HTTP_ORIGIN")
     end
 
+    def allow_headers
+      ALLOW_METHODS + @@additional_headers
```
raulb: s/ALLOW_METHODS/ALLOW_HEADERS

gudmundur:
Nicely done. I'm glad to see this contribution as the example you gave is something that's made my eyes all teary a few times. 😢

PR #322, opened 2017-12-06
Add Pliny::Helpers::ZuluTime to Serializers::Base

ransombriggs:
@gudmundur this adds a `ZuluTime` module to the pliny library that is mixed into `Serializers::Base` to provide a `zulu_time` method to help serialize dates properly.  One of the apps I am extracting this out of uses `DateTime` instances which do not have a `getutc` until `active_support/core_ext/numeric/time` is mixed in.  Should I add a check to see if `time` responds to `getutc` and throw a more helpful error if it is not defined?


gudmundur:
Nice work @ransombriggs. The only thing missing that I notice is updating the generators so that new serializers will use this rather than `#iso8601`. Once that's in I think this is good to go.

@brandur Do you have comments on this?

ransombriggs:
@gudmundur pushed up serializer template changes in https://github.com/interagent/pliny/pull/322/commits/4a30e492d261fd04869cb10a7b85bd29ce0bc489

PR #323, opened 2017-12-12
Add "rollbar.person_data" to rollbar scope

ransombriggs:
I was attempting to add the logged in user to my rollbar reports and setting the rack env data did not work.  It did when I used the default reporter and tracked the difference down to here https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/middleware/rack.rb#L33 .  This just duplicates that functionality in the pliny reporter.  The test is a little clunky in that it tests some rollbar functionality, but I wanted to make sure that the person data is picked up correctly.

### DIFF
--- a/spec/error_reporters/rollbar_spec.rb
+++ b/spec/error_reporters/rollbar_spec.rb
@@ -34,14 +34,20 @@ describe Pliny::ErrorReporters::Rollbar do
     end
 
     context "given a rack_env with person data" do
-      let(:rack_env) { { "rollbar.person_data" => { email: "foo@bar.com" } } }
+      let(:rack_env) do
+        { "rollbar.person_data" => {
+          id: SecureRandom.uuid,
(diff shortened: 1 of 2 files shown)


ransombriggs:
@mikehale extracted the logic from our app, does this look good?

spec/error_reporters/rollbar_spec.rb
```diff
@@ -33,6 +33,18 @@
       ))
     end
 
+    context "given a rack_env with person data" do
+      let(:rack_env) { { "rollbar.person_data" => { email: "foo@bar.com" } } }
```
mikehale: Maybe include `id` and `username` too to document all the things rollbar accepts as shown here: https://rollbar.com/docs/notifier/rollbar-gem/ ? Either that or maybe add an example in the readme.

gudmundur:
LGTM

PR #324, opened 2018-01-12
Sort by filename when generating INSERTs

fdr:
Previously, the only thing keeping the output of .sql generation was
the insert order of tuples happening to the recall order...something
that can easily de-synchronize with multiple developers that may be
writing migrations at the same time.

Instead of getting the tuples in any order, sort them by their
content.


pedro:
:+1:!!!

gudmundur:
@brandur I don't mind cutting a release now and possibly a few more in the future. But it would be nice to promote someone at Heroku to take over that role.

PR #326, opened 2018-01-16
Build against the most recent versions of Ruby

cyberdelia:
Build against the most recent versions of Ruby, including 2.5.


brandur:
Thanks @cyberdelia. LGTM.

PR #327, opened 2018-07-20
Use the current ruby version when generating a new project

dmathieu:


brandur:
Seems like an improvement! It does have a slight disadvantage in that if the local creator's Ruby version is outdated, then the new Pliny project will also be created on that outdated version of Ruby.

dmathieu:
It does have that disadvantage, yet.
For the record, rails [does the same thimg](https://github.com/rails/rails/blob/master/railties/lib/rails/generators/rails/app/templates/Gemfile.tt#L4).

brandur:
> For the record, rails does the same thimg.

Ah, fair enough.

gudmundur:
> Seems like an improvement! It does have a slight disadvantage in that if the local creator's Ruby version is outdated, then the new Pliny project will also be created on that outdated version of Ruby.

I'm assuming this can be checked in code before generating? I.e. if Ruby version is older than what we require, warn or err?

gudmundur:
@dmathieu 2.4.0 sounds about right.

dmathieu:
I've updated the PR to do just that.

dmathieu:
Sorry, I didn't see the tests were failing. I've fixed them, the last failing one seems to be an unrelated issue.

PR #328, opened 2020-01-15
Handle a JSON Array of params

stevenharman:
Back in #317 we started using `Sinatra::IndifferentHash` for Sinatra v2+, but didn't handle the case of a JSON array being passed as the root element. The older `#indifferent_params` from Sinatra v1 did handle this case. We've roughly re-created that here, though we're a little more permissive in that we allow for things that are Hash-like (via `#to_hash`) and Array-like (via `#to_ary`) rather than relying on type.

### DIFF
--- a/lib/pliny/helpers/params.rb
+++ b/lib/pliny/helpers/params.rb
@@ -21,9 +21,19 @@ module Pliny::Helpers
     def load_params(data)
       # Sinatra 1.x only supports the method. Sinatra 2.x only supports the class
       if defined?(Sinatra::IndifferentHash)
-        Sinatra::IndifferentHash[data]
+        indifferent_params_v2(data)
       else
         indifferent_params(data)
+      end
+    end
+
+    def indifferent_params_v2(data)
+      if data.respond_to?(:to_hash)
+        Sinatra::IndifferentHash[data.to_hash]
+      elsif data.respond_to?(:to_ary)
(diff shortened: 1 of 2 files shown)


lib/pliny/helpers/params.rb
```diff
@@ -21,10 +21,21 @@ def parse_body_params
...
+        indifferent_params_v2(data)
       else
         indifferent_params(data)
       end
     end
+
+    def indifferent_params_v2(data)
+      case data
+      when Hash
+        Sinatra::IndifferentHash[data.to_hash]
```
ransombriggs: @stevenharman I think the `to_hash` is redundant here because we have already verified it is a `Hash`

lib/pliny/helpers/params.rb
```diff
@@ -21,10 +21,21 @@ def parse_body_params
...
+      when Array
+        data.to_ary.map { |item| Sinatra::IndifferentHash[item] }
```
ransombriggs: @stevenharman I think the `to_ary` is redundant here because have verified it is an `Array`.  Could you also make this recursive to support arrays of arrays plus arrays of primitives, etc.

lib/pliny/helpers/params.rb
stevenharman: Doh! Good catch.

lib/pliny/helpers/params.rb
stevenharman: Done.

PR #330, opened 2020-01-16
Test cleanup and consistency

stevenharman:
While investigating an issue with the updater command I noticed a few things:

1. We were hand-rolling a bunch of temp files to hard-coded locations, and then leaving them around.
1. There were some Ruby warnings due to using deprecated methods
1. Inconsistency in how we delimit strings
1. Some other inconsistencies and such (as reported to me via Vim-Ale while I was fixing the above)

This PR cleans all of that up.

### DIFF
--- a/spec/commands/creator_spec.rb
+++ b/spec/commands/creator_spec.rb
@@ -1,5 +1,5 @@
-require 'pliny/commands/creator'
-require 'spec_helper'
+require "pliny/commands/creator"
+require "spec_helper"
 
 describe Pliny::Commands::Creator do
   before do
@@ -7,21 +7,23 @@ describe Pliny::Commands::Creator do
   end
 
   describe "#run!" do
-    before do
-      FileUtils.rm_rf("/tmp/plinytest")
-      FileUtils.mkdir_p("/tmp/plinytest")
-      Dir.chdir("/tmp/plinytest")
+    around do |example|
+      Dir.mktmpdir("plinytest-") do |dir|
+        Dir.chdir(dir) do
+          example.run
+        end
+      end
     end
 
     it "copies the template app over" do
       @gen.run!
-      assert File.exists?("./foobar")
-      assert File.exists?("./foobar/Gemfile")
+      assert File.exist?("./foobar")
+      assert File.exist?("./foobar/Gemfile")
     end
 
     it "deletes the .git from it" do
       @gen.run!
-      refute File.exists?("./foobar/.git")
+      refute File.exist?("./foobar/.git")
     end
 
     it "deletes the .erb files from it" do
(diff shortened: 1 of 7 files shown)


spec/integration_spec.rb
```diff
@@ -2,19 +2,24 @@
 
 describe "Pliny integration test" do
   before(:all) do
-    FileUtils.rm_rf("/tmp/plinytest")
-    FileUtils.mkdir_p("/tmp/plinytest")
-    Dir.chdir("/tmp/plinytest")
+    @original_dir = Dir.pwd
```
stevenharman: We need to capture this here, and change back to this dir in the `after(:all)` because Rspec doesn't provide a block mechanism for the `:all` hooks. But it will at least guarantee the `after(:all)` will run and thus CD us back to the right place.

spec/commands/creator_spec.rb
```diff
@@ -1,27 +1,29 @@
-require 'pliny/commands/creator'
-require 'spec_helper'
+require "pliny/commands/creator"
```
bensymonds: Is there any chance you could split the formatting/style changes out into separate commits in future? It would make seeing the actual real changes a lot easier.

spec/commands/updater_spec.rb
```diff
@@ -10,19 +10,21 @@
...
-    before do
-      FileUtils.rm_rf("/tmp/plinytest")
-      FileUtils.mkdir_p("/tmp/plinytest")
-      Dir.chdir("/tmp/plinytest")
-      File.open("/tmp/plinytest/Gemfile.lock", "w") do |f|
-        f.puts "    pliny (0.6.3)"
+    around do |example|
+      Dir.mktmpdir("plinytest-") do |dir|
+        Dir.chdir(dir) do
+          File.open("./Gemfile.lock", "w") do |f|
```
bensymonds: Not a bit deal really, but just wondering if you actually need the `./` bit here..? (Same comment applies to all the other instances in this PR.)

spec/integration_spec.rb
bensymonds: 👍

spec/commands/creator_spec.rb
stevenharman: Yes, I can be more careful about how I build the commits via `git add -p`. My tooling (Vim + Ale + standard.rb/rubocop) basically take care of the syntax stuff for me these days, so I've gotten a bit lax about it. I'll try to keep them discrete going forward.

spec/commands/updater_spec.rb
stevenharman: Probably not...? I did it in hopes of being explicit that we're talking about paths from right where we are. Which is the same thing that not having the dot should do, but I feel like sometimes that nuance gets lost when glancing at the code. Happy to drop the `./` if you prefer.

spec/commands/creator_spec.rb
bensymonds: Thanks 😄Yeah it does add a bit of faff when making the commits but hopefully we can agree it's worth it to keep the changes clear 👍

spec/commands/updater_spec.rb
bensymonds: Meh nah, I'm not that bothered about it and it sounds like a good reason. All good 👍

PR #331, opened 2020-01-21
Allow activesupport 5 or 6

stevenharman:
We have a very minimal dependency on active support - limited to the inflection extensions in a generator, and some time extensions in a test. So it seems safe to continue allowing 5, while opening up 6. We'll lock out 7 because it doesn't exist, and we don't know what, if any, 
breaking changes it will bring.


stevenharman:
> make travis run the tests against both major versions of activesupport

Excellent idea. I will see get that added.

PR #332, opened 2020-01-28
Ensure all strings with spaces are quoted in logs

bensymonds:
The current logic covers the case when the log data is already a `String` instance, but not when the data is some other object that, when interpolated into the log string, results in a string containing spaces.

The example that led me here is where an exception is logged and the exception's message contains spaces. The exception gets interpolated into the string and has `to_s` called on it. [`Exception#to_s`](https://ruby-doc.org/core-2.6.5/Exception.html#method-i-to_s) returns the message, so the message (with its spaces) go into the log line, but it's bypassed the logic that adds quotes.

See the commits for more details. I slipped in a few cleanups and added some tests too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 ### Changed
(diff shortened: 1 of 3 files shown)


bensymonds:
@stevenharman Thanks for the review. FYI I amended the last commit to add a changelog entry. I'm going to go ahead and merge this. I actually have another PR I want to open, but after that I'll be bugging you for a release 😄

PR #333, opened 2020-02-03
Replace newlines in log strings

bensymonds:
Pliny logging follows the ["logfmt" pattern](https://brandur.org/logfmt). logfmt is unclear on what should happen with newline characters, but given it's line-based it doesn't seem to make sense to include them in the body of the log line. I couldn't decide between removing, replacing with a space character, or escaping. In the end I went with escaping since it makes it clear to a viewer that a newline _was_ there.

This includes a small refactor - see the commits for details.


bensymonds:
@stevenharman If you're thinking about merging and releasing #338, maybe we could get this one in too?

PR #336, opened 2020-09-08
Support for Sinatra 2.1

beanieboi:
Sinatra 2.1 changed how helpers are loaded. They switched from include to prepend https://github.com/sinatra/sinatra/commit/2db247dd33bb29732a8e1d7e3a2172ee9e153efb

Going forward we need to depend on Sinatra 2.1 or higher.

### DIFF
--- a/lib/pliny/helpers/serialize.rb
+++ b/lib/pliny/helpers/serialize.rb
@@ -1,6 +1,6 @@
 module Pliny::Helpers
   module Serialize
-    def self.included(base)
+    def self.prepended(base)
       base.send :extend, ClassMethods
     end
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,18 +3,15 @@ dist: xenial
 services:
   - postgresql
 rvm:
+  - 2.7
   - 2.6
   - 2.5
   - 2.4
 env:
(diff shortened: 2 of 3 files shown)


lib/pliny/helpers/serialize.rb
```diff
@@ -1,6 +1,6 @@
 module Pliny::Helpers
   module Serialize
-    def self.included(base)
+    def self.prepended(base)
```
bensymonds, line 3: Could we just have _both_? Then we wouldn't have to require sinatra >=2.1.

beanieboi: We could for sure, I'd like to keep the maintenance cost low and cut old versions quickly. This change in Pliny is the only reason why we can't use Sinatra 2.1 there shouldn't be any other blockers for your project to move to Sinatra 2.1 with this new version of Pliny. if you don't want to move to Sinatra 2.1 you can stay on the current Pliny version without missing anything. We are not shipping new features in Pliny, it's basically in maintenance mode only at this point.

bensymonds:
I looked into this a bit and opened #337 as an alternative. My preference is for #337 because, even though as a breaking change it's more effort,  it's the long-term (and "correct", I think) fix. But I may be underestimating the impact of the breaking change.

PR #338, opened 2020-12-02
Return 400s on unparseable JSON POST payloads

kbarber:
When someone tries to parse bad request JSON, pliny right now will by default return a 500 via the web with an exception such as this logged:

```
MultiJson::ParseError: unexpected character (after foo.bar) at line 1, column 14 [parse.c:724]
```

Invalid JSON post bodies should instead return a 400, and the parse error somewhere in the body.

This PR wraps around the place we parse JSON and returns the appropriate pliny exception with a more helpful message attached to the response.

### DIFF
--- a/lib/pliny/helpers/params.rb
+++ b/lib/pliny/helpers/params.rb
@@ -8,9 +8,13 @@ module Pliny::Helpers
 
     def parse_body_params
       if request.media_type == "application/json"
-        p = load_params(MultiJson.decode(request.body.read))
+        begin
+          decoded = MultiJson.decode(request.body.read)
+        rescue MultiJson::ParseError => e
+          raise Pliny::Errors::BadRequest, e.message
+        end
         request.body.rewind
-        p
+        load_params(decoded)
       elsif request.form_data?
         load_params(request.POST)
       else
--- a/spec/helpers/params_spec.rb
+++ b/spec/helpers/params_spec.rb
@@ -34,4 +34,12 @@ describe Pliny::Helpers::Params do
     post "/", "<hello>world</hello>", {"CONTENT_TYPE" => "application/xml"}
     assert_equal "{}", last_response.body
   end
+
+  it "should throw bad request when receiving invalid json via post" do
+    err = assert_raises(Pliny::Errors::BadRequest) do
+      post "/", "{\"foo\"}", {"CONTENT_TYPE" => "application/json"}
+    end
+
+    assert_match /unexpected token/, err.message
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+### Fixed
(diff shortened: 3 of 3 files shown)


kbarber:
@beanieboi this is what I was thinking in regards to those 500's we were getting. Much easier than wrapping our code in multiple parse error rescues :-). /cc @stevenharman.

spec/helpers/params_spec.rb
```diff
@@ -34,4 +34,18 @@ def app
...
+
+  it 'should throw bad request when receiving invalid json via post' do
+    assert_raises(Pliny::Errors::BadRequest) do
+      post '/', '{\"foo\"}', {"CONTENT_TYPE" => "application/json"}
+    end
+  end
+
+  it 'should throw bad request and the correct message when post has invalid json' do
+    begin
+      post '/', '{\"foo\"}', {"CONTENT_TYPE" => "application/json"}
```
stevenharman, line 46: For consistency, could we change these to use double-quotes?

```suggestion
  it "should throw bad request when receiving invalid json via post" do
    assert_raises(Pliny::Errors::BadRequest) do
      post "/", "{\"foo\"}", {"CONTENT_TYPE" => "application/json"}
    end
  end

  it "should throw bad request and the correct message when post has invalid json" do
    begin
      post "/", "{\"foo\"}", {"CONTENT_TYPE" => "application/json"}
```

stevenharman (changes requested):
Nice find, and thank you for the fix. I do have a question about _maybe_ simplifying the code a little, and a nit re: style.

lib/pliny/helpers/params.rb
```diff
@@ -8,7 +8,12 @@ def body_params
...
       if request.media_type == "application/json"
-        p = load_params(MultiJson.decode(request.body.read))
+        begin
+          decoded = MultiJson.decode(request.body.read)
+        rescue MultiJson::ParseError => e
+          raise Pliny::Errors::BadRequest, e.message
+        end
+        p = load_params(decoded)
         request.body.rewind
         p
```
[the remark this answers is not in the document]
kbarber: No problemo.

spec/helpers/params_spec.rb
kbarber: Yeah no worries. My rubocop was defaulting back to "single quote style unless necessary".

kbarber:
@stevenharman all fixed, thanks for the comments.

spec/helpers/params_spec.rb
```diff
@@ -34,4 +34,18 @@ def app
...
+    end
+  end
+
+  it "should throw bad request and the correct message when post has invalid json" do
+    begin
+      post "/", "{\"foo\"}", {"CONTENT_TYPE" => "application/json"}
+    rescue Pliny::Errors::BadRequest => e
+      assert_match /unexpected token/, e.message
+    end
+  end
```
stevenharman, line 50: Looking at these a second time, I'm wondering if we could/should combine these into a single case. Thoughts?

```suggestion
  it "should throw bad request when receiving invalid json via post" do
    err = assert_raises(Pliny::Errors::BadRequest) do
      post "/", "{\"foo\"}", {"CONTENT_TYPE" => "application/json"}
    end

    assert_match /unexpected token/, err.message
  end
```

kbarber: Good thinking +1

stevenharman (approved):

PR #339, opened 2021-01-21
prepare for newer Ruby versions

beanieboi:
Testing newer ruby versions


lib/pliny/middleware/canonical_log_line.rb
```diff
@@ -58,9 +58,9 @@ class LogLine
       log_field :timing_serializer, Float
     end
 
-    def initialize(app, emitter:)
+    def initialize(app, options)
       @app = app
-      @emitter = emitter
+      @emitter = options[:emitter]
```
stevenharman, line 63: Since we _must_ have an `emitter`, maybe an `options.fetch(:emitter)` would be better?

beanieboi: I agree, updated

mathias (approved):
LGTM

stevenharman (approved):

PR #340, opened 2021-01-21
Allow newer version of Thor

beanieboi:
This allows newer versions of `Thor` and also moves some of the development dependencies.

We need a newer Thor version because some of our dependencies won't upgrade since Pliny is stuck on this old version


mathias (approved):
LGTM :shipit:

PR #342, opened 2021-06-17
Cut new release

stevenharman:
Cut a new release so we can get the relaxed Thor dependencies and thus upgrade our components.


mathias (approved):
LGTM!

beanieboi (approved):

PR #348, opened 2022-09-26
Support Sinatra 3

serprex:
Also test against Ruby 3.1. Drop testing against 2.5 as Sinatra 3 requires at least Ruby 2.6


beanieboi:
thank you, we did this in #356

PR #349, opened 2022-09-28
extend activesupport version range to include 7

serprex:
activesupport 7 requires Ruby >=2.7


beanieboi:
thank you, we did this in #357

PR #355, opened 2023-10-09
update supported rubies

beanieboi:
drop support for ruby 2.5, 2.6 and 2.7
add support for ruby 3.1 and 3.2

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby-version: ['2.5', '2.6', '2.7', '3.0']
+        ruby-version: ['3.0', '3.1', '3.2']
         activesupport-version: ['5', '6']
         sinatra-version: ['1', '2']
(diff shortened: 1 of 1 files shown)


dstarner (approved):

edmorley:
@beanieboi Hi! Should `gem.required_ruby_version` be set in `pliny.gemspec` too, to ensure newer pliny versions aren't incorrectly installed when using older Ruby?

Since Dependabot just opened a PR updating Cheverny to newer pliny, but CI still passed even though it's using Ruby 2.7 which is no longer tested in CI on pliny's side.

beanieboi:
> @beanieboi Hi! Should `gem.required_ruby_version` be set in `pliny.gemspec` too, to ensure newer pliny versions aren't incorrectly installed when using older Ruby?
> 
> Since Dependabot just opened a PR updating Cheverny to newer pliny, but CI still passed even though it's using Ruby 2.7 which is no longer tested in CI on pliny's side.

Yeah that's a good idea. I opened https://github.com/interagent/pliny/pull/365 for it

PR #357, opened 2023-10-09
update supported activesupport versions

beanieboi:
drop support for activesupport 5
add support for activesupport 7


spec/spec_helper.rb
```diff
@@ -13,6 +13,7 @@
 require "sinatra/namespace"
 require "sinatra/router"
 require "timecop"
+require "active_support/all"
```
tt, line 16: It looks like these were sorted alphabetically.

dstarner (approved):
just agreeing with @tt 's comment, but otherwise looks fine.

PR #370, opened 2025-09-04
misc maintenance work

beanieboi:
- sort dependencies in pliny.gemspec alphabetically
- default to sinatra 3 because thats the oldest version we support
- add ostruct as dependency, Ruby 3.5 will drop it as dependency
- unpin development dependencies
- Update templates to target Puma 7
- Add RuboCop config and GitHub Actions worklfow
- update bin/run template to work with newer RuboCop versions


jabrown85 (approved):
It looks good to me - but I don't think I have any pliny updates to upgrade so this is more a cursory review from my perspective 👍

PR #374, opened 2025-09-15
add sentry error reporter

brahyt-sf:
This is adding sentry to the available error reporters in pliny. This adds in similar person data extraction that is found in `error_reporters/rollbar.rb`

Testing to ensure duplicate rollbar and sentry output.

### DIFF
--- a/spec/error_reporters/sentry_spec.rb
+++ b/spec/error_reporters/sentry_spec.rb
@@ -43,7 +43,6 @@ describe Pliny::ErrorReporters::Sentry do
       expect(::Sentry).to have_received(:capture_exception).with(exception)
     end
 
-
     context "given a rack_env with sentry.person_data" do
       let(:rack_env) { { "sentry.person_data" => { id: 123, email: "test@example.com", username: "testuser" }, "rack.input" => StringIO.new } }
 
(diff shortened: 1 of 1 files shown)


brahyt-sf:
doh, sorry @beanieboi. I forgot to tag you into this. This was what I have been using for our work.

beanieboi:
The PR looks good. I'm going to merge it when I'm releasing a new version.

lib/pliny/error_reporters/sentry.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require "sentry-ruby"
```
tt, line 3: I think this should be required automatically by being declared in the Gemfile.

lib/pliny/error_reporters/sentry.rb
```diff
@@ -0,0 +1,46 @@
...
+          ::Sentry.capture_exception(e)
+        end
+      end
+
+      def extract_person_data_from_controller(env)
+        if env.key?("sentry.person_data")
+          env["sentry.person_data"] || {}
+        else
+          {}
+        end
```
tt, line 42: Is `env` a `Hash`?

If so, accessing an unknown key will already return `nil` so this can be simplified:

```suggestion
        env["sentry.person_data"] || {}
```

lib/template/config/initializers/sentry.rb
tt, line 1: I think we also want to remove the Rollbar initializer to not set up both for new projects.

tt (approved):

